PR #6, opened 2016-01-12
Allow either clang or LLVM to trigger builds

jfbastien:
This is required to support: https://codereview.chromium.org/1577943002/

Eventually we'll want other repositories to also be able to trigger builds. Maybe the lastest and lkgr files should also contain all the hashes that they've checked out, not just the one from BUILDBOT_REVISION. With that the build could use lkgr to only trigger on one repository, and keep all other repositories at lkgr (so they don't all float).

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -94,10 +94,17 @@ OCAML_BIN_DIR = os.path.join(OCAML_OUT_DIR, 'bin')
 
 NPROC = multiprocessing.cpu_count()
 
-# Try to use the LLVM revision provided by buildbot.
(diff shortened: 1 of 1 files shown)


jfbastien:
Note that this isn't ready to commit: https://github.com/WebAssembly/binaryen relies on latest / lkgr and I'll need to teach it about the new format.

I think I should change their output format to JSON too.

flagxor:
lgtm

jfbastien:
Following up in #8.

PR #12, opened 2016-01-14
Revert "Build V8"

jfbastien:
Reverts WebAssembly/waterfall#11

Gyp isn't available it looks like. I'll investigate and fix later.
```
  File "build/gyp_v8", line 35, in <module>
    import gyp_environment
  File "/b/build/slave/linux/build/src/src/work/v8/build/gyp_environment.py", line 13, in <module>
    import vs_toolchain
  File "/b/build/slave/linux/build/src/src/work/v8/build/vs_toolchain.py", line 24, in <module>
    import gyp
ImportError: No module named gyp
  File "/b/build/slave/linux/build/src/src/build.py", line 589, in <module>
    sys.exit(main())
  File "/b/build/slave/linux/build/src/src/build.py", line 566, in main
    BuildV8()
  File "/b/build/slave/linux/build/src/src/build.py", line 425, in BuildV8
    proc.check_call(['build/gyp_v8'], cwd=V8_SRC_DIR)
  File "/b/build/slave/linux/build/src/src/proc.py", line 34, in check_call
    return subprocess.check_call(cmd, **kwargs)
  File "/usr/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['build/gyp_v8']' returned non-zero exit status 1
```

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -563,7 +563,6 @@ def main():
   BuildLLVM()
   TestLLVM()
   InstallLLVM()
-  BuildV8()
   BuildSexpr()
   BuildOCaml()
   BuildSpec()


binji:
You probably need to run `gclient sync`.

jfbastien:
@binji I was just talking about that with @dschuff: I don't think I can do that from the bots because `gclient` leaves the network and can cause flakes. So far the waterfall avoids using non-mirror things by always using `https://chromium.googlesource.com/` (even for github repos) and `https://storage.googleapis.com/`.

Or does `gclient` avoid this? /cc @flagxor

jfbastien:
Ha, actually the bot [already does this](https://build.chromium.org/p/client.wasm.llvm/builders/linux/builds/1917/steps/steps/logs/stdio):
```
/b/build/../depot_tools/gclient sync --force --verbose --jobs=2 --output-json /tmp/tmpDjBSK9
```
So presumably it's OK to do for sub-repos as well.

dschuff:
We use gclient this way in all of our builds already, for Chromium and
nacl, so I don't think it will be a problem.

jfbastien:
OK, trying in #13.

jfbastien:
That didn't work, same `import gyp` failure. Reverting for now, it's probably just missing from the Python search path.

PR #15, opened 2016-01-15
Add utility to execute programs.

jfbastien:
I'll need to change binaryen's binaryen-shell to support the -o option.


jfbastien:
@kripken as discussed yesterday, I think something like this would work for the waterfall as well as for `check.py`. I just need to move `binaryen-shell` to use `support/command-line.h`, I;ll do it this afternoon.

kripken:
I don't exactly see how these pieces fit together (would this be called from check.py?), but in general this all sounds fine, no concerns.

PR #19, opened 2016-01-26
Add options to skip sync and/or build stages

dschuff:
This makes it much easier to run the integration tests by themselves.
Also disable the V8 lint checks; they depend on depot_tools and don't
work everywhere.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -14,7 +14,6 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-import argparse
 import glob
 import json
 import multiprocessing
@@ -445,8 +444,8 @@ def SyncRepos():
 
 
 def GetRepoInfo():
-  # Keep track of all repo information here, preventing the summary from
-  # getting out of sync with the actual list of repos.
+  """Collect a readable form of all repo information here, preventing the
+  summary from getting out of sync with the actual list of repos."""
   info = {}
   for r in ALL_SOURCES:
     info[r.name] = r.CurrentGitInfo()
@@ -651,6 +650,7 @@ def Summary(repos):
 
 
 def ParseArgs():
+  import argparse
   parser = argparse.ArgumentParser(
       description='Wasm waterfall top-level CI script')
   parser.add_argument('--no-sync', dest='sync',
--- a/src/build.py
+++ b/src/build.py
@@ -14,6 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import argparse
 import glob
 import json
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -625,19 +626,36 @@ def Summary(repos):
...
+                      default=True, action='store_false',
+                      help='Skip fetching and checking out source repos')
+  parser.add_argument(
+      '--no-build', dest='build', default=True, action='store_false',
+      help='Skip building source repos (also skips V8 and LLVM unit tests)')
+  return parser.parse_args()
+
+
 def main():
+  options = ParseArgs()
```
jfbastien: Could you instead follow the approach other `.py` files do in this repo: call `ParseArgs` only when not `__main__`? Right now it works because you have all-defaults, but won't work otherwise.

src/build.py
```diff
@@ -625,19 +626,36 @@ def Summary(repos):
...
-  Remove(INSTALL_DIR)
-  repos = SyncRepos()
-  LLVM()
-  V8()
-  Sexpr()
-  OCaml()
-  Spec()
-  Binaryen()
-  ArchiveBinaries()
+  repos = {} if not options.sync else SyncRepos()
```
jfbastien: `repos` isn't used?

jfbastien: I'd still generate content for `repos` by splitting out the `Sync` part from the one that goes through and figures out where each repo is. That gets printed out by `Summary` and makes it easier to figure out WTH was tested.

src/build.py

src/build.py
dschuff: OK, you asked for it...

src/build.py
```diff
@@ -401,30 +438,18 @@ def Clobber():
...
-  ]
+  for repo in ALL_SOURCES:
+    repo.Sync()
   SyncLLVMClang()
   SyncOCaml()
+
+
+def GetRepoInfo():
   # Keep track of all repo information here, preventing the summary from
   # getting out of sync with the actual list of repos.
```
jfbastien: Change to """string comment"""

src/build.py
```diff
@@ -625,19 +650,37 @@ def Summary(repos):
     UploadToCloud('lkgr', 'git/lkgr', 'lkgr')
 
 
-def main():
+def ParseArgs():
+  parser = argparse.ArgumentParser(
```
jfbastien: import argparse here instead of at the top.

jfbastien:
lgtm besides nit. Please wait for the waterfall to be green before merging to avoid piling problems.

PR #20, opened 2016-01-26
Limit testing commands to 10 CPU seconds.

dschuff:
Because I may or may not have written an infinite loop into LLVM at some
point in my development...

### DIFF
--- a/src/testing.py
+++ b/src/testing.py
@@ -19,6 +19,7 @@ import math
 import multiprocessing
 import os
 import os.path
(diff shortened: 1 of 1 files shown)


dschuff:
This seems uncontroversial, just landing it.

jfbastien:
Ah nice, ty. What does it say when it fails?

dschuff:
clang-3.9: error: unable to execute command: Killed

PR #22, opened 2016-01-31
Add new d8 failure

jfbastien:


jfbastien:
FYI @binji @titzer this looks like a binary format mismatch?

PR #25, opened 2016-02-06
Mark some d8 failures

jfbastien:
It looks like this fix didn't quite work:
https://codereview.chromium.org/1624323003/

Or something else is mismatching. In any case, mark as broken for now because I'm about to board a plane.


jfbastien:
FYI @titzer @dschuff I didn't investigate the issue, but something's broken. Let's fix on Monday.

PR #26, opened 2016-02-14
Fast-forward when not on master.

jfbastien:


jfbastien:
@dschuff I think this is right? Otherwise non-master branches lag behind (e.g. musl). I'll merge this in, LMK if my git-fu is erring.

dschuff:
For update workflows designed for bots, we should not even have local branches (or in other words the workflow should not depend on their existence or nonexistence). So if we just want the branch head instead of a specific revision we should do `git checkout origin/<branchname>` for master or any other branch, and then in all cases that would result in a detached head with the right version, and bots would never care about local branches (practically this would mean there would be an out-of-date master branch because it's there by default, and no other branches). Locally you can keep local branches up to date or do whatever you want. I just usually use `--no-sync` and sync things manually.

PR #27, opened 2016-02-14
Fix fast-forward.

jfbastien:


jfbastien:
@dschuff this also handled detached HEAD case.

PR #29, opened 2016-02-23
Update spec expected failures

jfbastien:
spec now understands abort / exit, leading to multiple new passes (a few still fail because abort/exit were hiding other imports). The next step will be to make binaryen generate (start ...) so that the spec repo executes the content (which will create new failures).


jfbastien:
FYI @rossberg-chromium

PR #30, opened 2016-02-24
Work around spec issue

jfbastien:
https://github.com/WebAssembly/spec/issues/248 is causing the waterfall to fail. Work around the issue.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -261,6 +261,10 @@ class Source:
     """Clone a git repo if not already cloned, then fetch and checkout."""
     if os.path.isdir(self.src_dir):
       print '%s directory already exists' % self.name
+      if self.name == 'spec':
+        # TODO(jfb): Work around https://github.com/WebAssembly/spec/issues/248
+        proc.check_call(['git', 'checkout', '--', 'ml-proto/winmake.bat'],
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -261,6 +261,10 @@ def GitCloneFetchCheckout(self):
     """Clone a git repo if not already cloned, then fetch and checkout."""
     if os.path.isdir(self.src_dir):
       print '%s directory already exists' % self.name
+      if self.name == 'spec':
+        # TODO(jfb): Work around https://github.com/WebAssembly/spec/issues/248
+        proc.check_call(['git', 'checkout', '--', 'ml-proto/winmake.bat'],
```
dschuff: How about just git reset --hard?

jfbastien: I don't want to clobber local changes... well, except for this one ;-)

dschuff: Or better yet:
git clean -dffx
git reset --hard HEAD

see https://chromium.googlesource.com/native_client/src/native_client/+/master/pynacl/repo_tools.py#172

jfbastien: I just want to work around that one thing, without resetting everything else.

PR #31, opened 2016-02-25
s2wasm: add start

jfbastien:


jfbastien:
FYI, I'm going to do this, but it's not ready yet because V8 is sad:
https://codereview.chromium.org/1734533004/

I'll fix the V8 bug before merging this.

PR #34, opened 2016-02-29
Work around sexpr-wasm dirty repo

jfbastien:
Work around https://github.com/WebAssembly/sexpr-wasm-prototype/issues/35 because it makes the waterfall sad.


jfbastien:
FYI @binji I'll work around this for now, since it's pretty quick and the waterfall is sad.

PR #36, opened 2016-03-01
Build fastcomp and emscripten

dschuff:
Build rules for fastcomp and emscripten.


dschuff:
Should we switch to the incoming branch?

jfbastien:
Yes, incoming branch would be better. /cc @kripken

kripken:
If the goal is to track development as it happens, instead of with a few weeks delay, then incoming is best.

PR #37, opened 2016-03-02
Build torture tests with emcc/asm2wasm

dschuff:


src/build.py
```diff
@@ -592,6 +596,7 @@ def Binaryen():
     f = os.path.join(BINARYEN_BIN_DIR, node)
     if os.path.isfile(f):
       CopyBinaryToArchive(f)
+  CopyBinaryToArchive(os.path.join(BINARYEN_SRC_DIR, 'bin', 'wasm.js'))
```
jfbastien: Hmm, that's a bit confusing since we also have one in /lib :s

src/build.py
```diff
@@ -620,17 +625,26 @@ def Fastcomp():
...
-  os.environ['LLVM'] = os.path.join(INSTALL_DIR, 'fastcomp', 'bin')
-  os.environ['V8'] = os.path.join(INSTALL_DIR, 'bin', 'd8')
   em_config = os.path.join(INSTALL_DIR, 'emscripten_config')
+  emscripten_dir = os.path.join(INSTALL_DIR, 'emscripten')
+  os.environ['EM_CONFIG'] = em_config
+  Remove(emscripten_dir)
+  shutil.copytree(EMSCRIPTEN_SRC_DIR,
+                  emscripten_dir,
+                  symlinks=True,
+                  ignore=shutil.ignore_patterns('.git'))
```
jfbastien: Why? (add comment)

src/build.py
```diff
@@ -620,17 +625,26 @@ def Fastcomp():
...
   shutil.copy2(os.path.join(SCRIPT_DIR, 'emscripten_config'),
                os.path.join(em_config))
   try:
     proc.check_call([
-        os.path.join(EMSCRIPTEN_SRC_DIR, 'emcc'),
-        '--em-config', em_config,
+        os.path.join(emscripten_dir, 'emcc'),
```
jfbastien: Could you move emcc to INSTALL_DIR?

src/build.py
dschuff: Yup. we could rename one or both of them in their respective repos, but for now waterfall just needs to put them where stuff expects them.

src/build.py
dschuff: Emscripten's source checkout is also its distribution package. But we don't want to archive that big blob. Added a comment.

src/build.py
dschuff: I suppose I could. How do you feel about having the `bin/` dir of your package polluted by 100 random source files?
I mean it's not really a distribution package that we're making, so it kind of doesn't matter either way I guess.

jfbastien: hrm... I think we do want to distribute it in the tarball, but avoiding pollution / bloat / overriding. Maybe emscripten could be in a separate top-level dir? Or rather, what's the minimum that's needed to emscripten can do its thing?

dschuff: Oh and actually that's kind of a pain to do because shutil.copytree requires that the destination dir not exist. So you basically have to reimplement cp -r.

Which I guess I already did in PNaCl's build system so whatever.

jfbastien: If it's in a subdir (because it pollutes) then copytree will work. It's ugly, but less pollution?

kripken:
Btw, I've been working towards doing the same in binaryen. Testing in both places is useful, I think.

dschuff:
Agreed, and having different tests is useful too.

src/build.py
dschuff: True, but is having `bin/emscripten/emcc` any less ugly than having `emscripten/emcc` ?
Maybe so. I'll do that.

jfbastien:
lgtm with one or the other approach :)

PR #38, opened 2016-03-03
Run asm2wasm-generated tests in v8

dschuff:
Add binaryen to the emcc build of the torture tests.

Currently Binaryen's output only supports interpreting the wast files
with wasm.js (running in the underlying JS environment doesn't work
yet). So we just test that for now. It also expects all the
files (eg. the wast file associated with the JS file) to be in the cwd,
so change the support for setting the outdir in
ExecuteLLVMTorture (remove the unused has_output arg, and use an
expclicit out dir instead), and fix up some plumbing to allow that.


jfbastien:
lgtm

PR #40, opened 2016-03-17
Switch the asm2wasm tests to use V8's native wasm support

dschuff:
Previously the asm2wasm tests used the binaryen interpreter compiled
into wasm.js. Now we use Binaryen's native-wasm method to use V8
directly. However we still use sexpr-wasm to encode the binary, so that
Binaryen's binary format need not match V8's.

### DIFF
--- a/src/compile_torture_tests.py
+++ b/src/compile_torture_tests.py
@@ -26,7 +26,12 @@ CFLAGS_COMMON = ['--std=gnu89', '-DSTACK_SIZE=1044480',
                  '-w', '-Wno-implicit-function-declaration']
 CFLAGS_EXTRA = {
     'wasm': ['--target=wasm32-unknown-unknown', '-S', '-O2'],
-    'asm2wasm': ['-s', 'BINARYEN=1', '-s', 'BINARYEN_METHOD="wasm-s-parser"'],
+    # Binaryen's native-wasm method uses the JS engine's native support for
+    # wasm rather than interpreting the wasm with wasm.js.
+    # There is also 'wasm-binary'; it uses binaryen's binary format, which
+    # may not match v8's format exactly. So we just generate the wast with
+    # emcc and use sexpr-wasm to generate the binary.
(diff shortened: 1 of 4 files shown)


kripken:
Note btw that binaryen's binary format should now match the spec, except for being postorder. So hopefully we'll soon have an identical binary format between binaryen and sexpr-wasm.

dschuff:
Cool! Although I also anticipate that not all the binary format changes (and other behavior changes as the spec evolves) will get implemented at exactly the same time by all of the relevant codebases (V8, SM, Binaryen, sexpr-wasm, spidermonkify), so there will necessarily be some skew. However I do expect that V8 and sexpr-wasm will be very closely coupled, so I may keep it this way to insulate the waterfall from some of that until things stabilize more.

dschuff:
By the way, here's my interpretation of the binaryen methods, you can tell me if I'm correct: `native-wasm` loads the `.wast` directly into the JS engine (which is supported by SM but not V8), and `wasm-binary` creates a binary `.wasm` with Binaryen's encoder, and interprets that with `wasm.js`. What happens when you compose those 2 in a comma-separated list?

kripken:
Yeah, I think you're right that skew will happen. Good to decouple.

Hmm, we should rename those binaryen methods, now that you mention it. `wasm-binary` should be `interpret-binary` (or `polyfill-binary`?). So `native-wasm,interpret-binary` should try native wasm support, and if not present, run the binary in the interpreter. The opposite, `interpret-binary,native-wasm` would try the interpreter first, but it will always work, so the second method is unneeded.

kripken:
`native-wasm` should also use a binary file, yeah. But why do you say it uses a wast? The code does `instantiateModule` which should work in all browsers?

dschuff:
I think if you just use `native-wasm` emscripten will not produce a binary file, it will only produce a wast file.

kripken:
Ok, that sounds familiar now. Yeah, it used to be that way (and we depended on a script like spidermonkify), but I fixed it.

PR #41, opened 2016-03-21
Add submodule support to GitCloneFetchCheckout

dschuff:
sexpr-wasm won't build anymore (actually CMake won't configure) without
gtest. So we need to use at least that submodule.

So this CL adds submodule init/update suport to the sync. Instead of
using all submodules (which would also sync v8, which we don't need and
is large), pass an explicit list of submodules to init.

Also don't clobber the source directory anymore.


dschuff:
+cc @binji 
Maybe want to make the tests optional with a (maybe default-on) CMake flag or something?

binji:
Ah, good point. I'll add a CMake option to disable it.

PR #42, opened 2016-03-23
Display the list of sync and build targets in help

binji:
This looks like:
```
sync targets:
  binaryen chromium-clang clang emscripten fastcomp fastcomp-clang gcc llvm
  llvm-test-suite musl sexpr spec v8 waterfall

build targets:
  archive binaryen emscripten fastcomp llvm musl ocaml sexpr spec v8
```

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -829,14 +829,28 @@ def BuildRepos(filter=None):
 
 def ParseArgs():
   import argparse
+  import textwrap
 
   def SplitComma(arg):
     if not arg:
       return None
     return arg.split(',')
 
+  def TextWrapNameList(prefix, items):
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -21,6 +21,7 @@
 import os
 import shutil
 import sys
+import textwrap
```
dschuff: I suppose since JF insisted we defer/avoid import of argparse into `ParseArgs()` that we ought to do the same with textwrap.

dschuff:
Otherwise lgtm

src/build.py
binji: done.

PR #43, opened 2016-04-04
Add emscripten vanilla configuration

dschuff:
This change adds the 'vanilla' (aka wasm-backend) emscripten
configuration to the waterfall, and compiles and runs the torture tests
with it (in addition to asm2wasm).

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -70,7 +70,8 @@ BINARYEN_BIN_DIR = os.path.join(BINARYEN_OUT_DIR, 'bin')
 FASTCOMP_OUT_DIR = os.path.join(WORK_DIR, 'fastcomp-out')
 MUSL_OUT_DIR = os.path.join(WORK_DIR, 'musl-out')
 TORTURE_S_OUT_DIR = os.path.join(WORK_DIR, 'torture-s')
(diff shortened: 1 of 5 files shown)


src/build.py
```diff
@@ -71,6 +71,7 @@
 MUSL_OUT_DIR = os.path.join(WORK_DIR, 'musl-out')
 TORTURE_S_OUT_DIR = os.path.join(WORK_DIR, 'torture-s')
 ASM2WASM_TORTURE_S_OUT_DIR = os.path.join(WORK_DIR, 'torture-s-asm2wasm')
+EMSCRIPTENWASM_TORTURE_S_OUT_DIR = os.path.join(WORK_DIR, 'torture-s-emwasm')
```
jfbastien: `-s` is for `.s` files.

src/build.py
```diff
@@ -632,27 +637,33 @@ def Fastcomp():
 
 def Emscripten():
   buildbot.Step('emscripten')
+  Remove(os.path.expanduser(os.path.join('~', '.emscripten_cache')))
```
jfbastien: Comment why.

jfbastien:
lgtm otherwise

src/build.py
dschuff: Replaced with `asm2wasm-torture-out` et al.

src/build.py

PR #44, opened 2016-04-15
Archive emcc+wasm files

eholk:
It would be helpful to be able to download these files from each build. I'm using them for my own testing.


eholk:
Thanks!

PR #45, opened 2016-04-15
Set maximum cpu time to 90 seconds, as 10 seconds was too short

kripken:
(and hit some thread-sanitizer tests, see https://github.com/WebAssembly/binaryen/pull/348)


jfbastien:
lgtm, we could make it configurable at some later point :)

PR #46, opened 2016-05-05
Update expected failures after LLVM r268679

dschuff:
This seems to match what the waterfall is seeing; although I'm actually getting more new passes locally...


dschuff:
we'll see what happens to the waterfall

PR #54, opened 2016-07-12
Use ccache if present for building LLVM

dschuff:
Now that LLVM requires a recent CMake, this is easy to do.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -576,26 +576,42 @@ def GetRepoInfo():
   return info
 
 
+def Which(name):
+  """Find an executable on the system by name. If not found return ''."""
+  # If we want to run this on Windows, we'll have to be smarter.
+  try:
+    o = proc.check_output(['which', name])
+    return o.strip()
+  except proc.CalledProcessError:
+    return ''
+
+
 def LLVM():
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -576,26 +576,42 @@ def GetRepoInfo():
   return info
 
 
+def Which(name):
+  """Find an executable on the system by name. If not found return ''."""
+  # If we want to run this on Windows, we'll have to be smarter.
```
dekimir: SPIRV-Tools solved it like this: https://github.com/KhronosGroup/SPIRV-Tools/pull/132

dschuff: Hm interesting, I thought check_output worked on Windows. In any case the comment refers to the existence of the `which` program. Windows has a similar thing called `where` although for NaCl we did it manually: https://chromium.googlesource.com/native_client/src/native_client/+/master/pynacl/file_tools.py#72

dekimir: Oh, sorry, I forgot about `which` vs. `where`, so I thought you were talking about `check_output`.  Reportedly, certain minor versions of Python 2.7 don't support `check_output` on Windows.

PR #56, opened 2016-07-13
binaryen-shell has been renamed wasm-shell in binaryen

kripken:
Should be landed in sync with https://github.com/WebAssembly/binaryen/pull/632


jfbastien:
lgtm

PR #57, opened 2016-07-18
Make emscripten failures real and standalone failures warnings

dschuff:
Previously failures in the standalone build (without emscripten) caused
the build to be red, whereas failures with emscripten were just warnings
because the support was more experimental. This PR reverses that,
because emscripten is what we expect users to be using, and the
standalone support is experimental. We keep the standalone builds
running because in the past they have sometimes been useful for catching
issues in implmentations (wasm-shell, v8, etc) and the maintenance cost
for keeping them running has been low.

Also update current failure state of tests


dschuff:
Hm, Travis doesn't seem to be running automatically. Will have to look into that

PR #59, opened 2016-08-04
Add support for running emscripten test suite

dschuff:
Run the emscripten test suite with the wasm backend, using the
wasm-shell (as opposed to native wasm) for now.

Also remove the attempt to use a relative path in the emscripten config
file and instead generate one with an absolute path at install time.


dschuff:
Oops, got the wrong description.

PR #60, opened 2016-08-04
Scrub PYTHONPATH from environment when running under buildbot

dschuff:
Chrome's buildbot infrastructure has a module named 'tools' which is
included in the PYTHONPATH by default and conflicts with emscripten's
own 'tools' module. The waterfall script doesn't use Chrome's 'tools',
so this change scrubs PYTHONPATH when running under buildbot.


dschuff:
@kripken with this patch the emscripten test suite is running with wasm backend on the waterfall \o/

PR #61, opened 2016-08-09
Make build.py use the prebuilt cmake binary for all builds which use it.

jgravelle-google:
I ran in to issues getting my build set up. On further investigation it makes sense to me that we use a single consistent version of cmake across all builds, rather than having a prebuilt bin and the bin that the computer this runs on happens to have.


dschuff:
LGTM; the build is hosed right now for unrelated reasons; I just checked in rev 57b2296 to fix that so once a build runs from that rev, I'll merge this one.

PR #62, opened 2016-08-18
Binaryen 0xb support

jgravelle-google:
Upstream binaryen has started using 0xc features. sexpr-wasm doesn't have support for that yet, so for now waterfall will sync and build both versions of binaryen.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -15,6 +15,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import errno
 import glob
 import json
 import multiprocessing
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -693,27 +709,37 @@ def Spec():
...
        '-DCMAKE_CXX_COMPILER=' + CXX],
-      cwd=BINARYEN_OUT_DIR)
-  proc.check_call(['ninja'], cwd=BINARYEN_OUT_DIR)
-  assert os.path.isdir(BINARYEN_BIN_DIR), 'Expected %s' % BINARYEN_BIN_DIR
-  for node in os.listdir(BINARYEN_BIN_DIR):
-    f = os.path.join(BINARYEN_BIN_DIR, node)
+      cwd=out_dir)
+  proc.check_call(['ninja'], cwd=out_dir)
+  bin_dir = os.path.join(out_dir, 'bin')
+  assert os.path.isdir(bin_dir), 'Expected %s' % bin_dir
```
dschuff: Should we bail here if archive_dir is None?

jgravelle-google: Ah, I missed a crucial piece that makes it work while trying to clean up the commit history.

Basically if it's None it gets ignored when we call CopyBinaryToArchive

... I also missed the piece that actually *uses* archive_dir in the call to CopyBinaryToArchive.

dschuff:
LGTM

PR #63, opened 2016-08-19
Use 0xb version of wasm-shell for standalone tests and interpret-binary for emscripten tests

dschuff:
This test executes the code generated by 0xb s2wasm, so it should be the
0xb interpreter.


dschuff:
Actually hold off, we need one more thing...

dschuff:
OK, PTAL

jgravelle-google:
LGTM

PR #64, opened 2016-08-23
Go back to the LLVM github mirror for LLVM and clang

dschuff:


dschuff:
I notice you committed these; is this change ready?

src/build.py
```diff
@@ -343,6 +344,8 @@ def GitCloneFetchCheckout(self):
     """Clone a git repo if not already cloned, then fetch and checkout."""
     if os.path.isdir(self.src_dir):
       print '%s directory already exists' % self.name
+      proc.check_call(['git', 'remote', 'set-url', 'origin', self.git_repo],
```
dschuff: Actually, I'm not sure we want this. This will reset the `origin` remote even for local users. If we're going to be clever here, we should be more clever and if `origin` is the `external/llvm.org` mirror, then reset it, but otherwise don't touch it.

jgravelle-google: Right, the intent was that if you're not doing a clean fresh-from-upstream build, that if we changed the origin in the script, this would update it for you. On reflection that might not be desired if the user had overridden origin intentionally.

Possibly we could set up a new remote, say 'waterfall', that the script uses and overwrites, and local users are free to manually sync an unchanging origin.

dschuff: I like that idea. I think `git clone` will create the `origin` remote by default, but we can also add a `waterfall` remote that can duplicate it (or not, if the user changes `origin`). Then in the places where we hardcode `origin/master` as the revision we check out now, we can change those to `waterfall`.

jfbastien: `wasm-stat.us` as a remote name will be less likely to collide :-)

PR #65, opened 2016-08-23
Fix waterfall remote setup for v8 and waterfall

dschuff:
v8 and the waterfall repo itself are special snowflakes.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -332,6 +332,17 @@ def RemoteBranch(branch):
   return WATERFALL_REMOTE + '/' + branch
 
 
+def GitUpdateRemote(src_dir, git_repo, remote_name):
+  try:
+    proc.check_call(['git', 'remote', 'set-url', remote_name, git_repo],
+                    cwd=src_dir)
+  except proc.CalledProcessError:
+    # If proc.check_call fails it throws an exception. 'git remote set-url'
+    # fails when the remote doesn't exist, so we should try to add it.
+    proc.check_call(['git', 'remote', 'add', remote_name, git_repo],
+                    cwd=src_dir)
+
+
 class Source:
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -388,9 +389,13 @@ def pretty(fmt):
       return proc.check_output(
           ['git', 'log', '-n1', '--pretty=format:%s' % fmt],
           cwd=self.src_dir).strip()
-    remote_name = 'remote.%s.url' % WATERFALL_REMOTE
-    remote = proc.check_output(['git', 'config', '--get', remote_name],
-                               cwd=self.src_dir).strip()
+    try:
+      remote = GitRemoteUrl(self.src_dir, WATERFALL_REMOTE)
+    except proc.CalledProcessError:
+      # Not all checkouts have the '_waterfall' remote (e.g. the waterfall
```
jgravelle-google: The waterfall syncs itself? I thought its custom_sync was set to NoSync meaning it didn't sync?

src/build.py
```diff
@@ -332,6 +332,17 @@ def RemoteBranch(branch):
   return WATERFALL_REMOTE + '/' + branch
 
 
+def GitUpdateRemote(src_dir, git_repo, remote_name):
```
jgravelle-google: This really does belong more as a top-level construct. I had felt it odd to have a function that took three arguments where two of which were member variables when it was only used inside that class.
So, nice!

jgravelle-google:
LGTM

src/build.py
dschuff: It doesn't sync itself, so there won't be a `_waterfall` remote, but this code (to print the git info) runs on both the waterfall and its sub-checkouts.

PR #67, opened 2016-09-01
Add prebuilt cmake to PATH

jgravelle-google:
Some emscripten tests use the system's cmake, which doesn't work on the
build bots. Add our prebuilt version of cmake at the start of PATH to
guarantee consistency in any subprocesses that rely on cmake.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -466,10 +466,6 @@ def SyncPrebuiltCMake(name, src_dir, git_repo):
       print 'Error downloading %s: %s' % (url, e)
       raise
 
-  # Add prebuilt cmake to PATH so any subprocesses use a consistent cmake.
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -466,6 +466,10 @@ def SyncPrebuiltCMake(name, src_dir, git_repo):
       print 'Error downloading %s: %s' % (url, e)
       raise
 
+  # Add prebuilt cmake to PATH so any subprocesses use a consistent cmake.
+  os.environ['PATH'] = (os.path.join(PREBUILT_CMAKE_DIR, 'bin') +
+                        ':' + os.environ['PATH'])
```
dschuff: let's make the ':' `os.pathsep` so there's at least a chance that it will work on windows.

jgravelle-google: Makes sense - did not know that was a thing

dschuff:
Lgtm

PR #71, opened 2016-09-13
Use gn to generate V8 ninja files

### LINKED ISSUE #69: V8 build needs to switch to GN
V8's gyp build is now deprecated (or almost-deprecated; slightly less deprecated than the gyp build of Chrome) and V8's bots are now using GN. We should switch too. Instructions are at https://github.com/v8/v8/wiki/Building-with-GN

dschuff:
Fixes #69.
It turns out that the gyp generator no longer even runs from the gclient
hook, so we were just going on ninja files generated from old revisions.


jgravelle-google:
lgtm

PR #73, opened 2016-09-15
In the case of an exception always mark current step as failing.

sbc100:
Without this change steps that throw exception are
erroneously marked as succeeding.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -24,6 +24,7 @@ import shutil
 import sys
 import tarfile
 import tempfile
+import traceback
 import urllib2
 
 import assemble_files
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -633,7 +634,7 @@ def SyncRepos(filter=None, sync_lkgr=False):
   buildbot.Step('Sync Repos')
 
   good_hashes = None
-  if options.sync_lkgr:
```
jgravelle-google: Huh. This technically works but is gross, right? Wouldn't have expected it to work, learned something today about Python scope resolution.

Thanks!

sbc100: Yes, globals in python are not always intuitive

PR #74, opened 2016-09-16
Add support for goma compiler when building LLVM

sbc100:


src/build.py
```diff
@@ -691,17 +691,28 @@ def LLVM():
...
                       c for c in ['C', 'CXX']])
 
+  if compiler_launcher:
+    command.extend(['-DCMAKE_%s_COMPILER_LAUNCHER=%s' %
+                    (c, compiler_launcher) for c in ['C', 'CXX']])
+
   proc.check_call(command, cwd=LLVM_OUT_DIR)
-  proc.check_call(['ninja', '-v'], cwd=LLVM_OUT_DIR)
+  proc.check_call(['ninja', '-v'] + jobs, cwd=LLVM_OUT_DIR)
+  sys.exit(1)
```
dschuff: I assume this was for local testing? :)

kripken:
Just curious, what's the goma compiler?

src/build.py
sbc100: Doh! Thanks

dschuff:
It's a distributed compilation infrastructure used by Chromium's buildbots. Basically it's just a big distcc that runs on all 3 OSes. Unfortunately I don't think it's available outside of Google, so we still have the ccache fallback (and of course the non-cached fallback).

sbc100:
goma is the distributed compiler used by chromium that performs compiles on GCE and has aggressive caching.  Right now its only available to google employees, but its also available on the builtbots which means it can be utilized there to speed up the build.  I'll update the CL description.

src/build.py
```diff
@@ -691,17 +691,27 @@ def LLVM():
...
              '-DLLVM_TARGETS_TO_BUILD=X86']
-  ccache = Which('ccache')
-  if ccache:
-      command.extend(['-DCMAKE_%s_COMPILER_LAUNCHER=%s' %
-                      (c, ccache) for c in ['C', 'CXX']])
+
+  compiler_launcher = None
+  jobs = []
+  if 'GOMA_DIR' in os.environ:
+    compiler_launcher = os.path.join(os.environ['GOMA_DIR'], 'gomacc')
```
dschuff: Do we need to run `goma_ctl.py ensure_start` or some such, or does the bot infra do that?

kripken:
Interesting, thanks for the info.

src/build.py
dschuff: Anyway I'm about to be gone for the day so LGTM if it's unnecessary or if it is necessary and you add it :)

sbc100: No, I think the bot setup takes care of all that.

dschuff: Hm, maybe not? https://wasm-stat.us/builders/linux/builds/11019/steps/LLVM/logs/stdio

sbc100: Arg.  Will fix.  Want me to revert this for now?

PR #75, opened 2016-09-16
Make torture tests work from any working directory

jgravelle-google:
Torture tests wouldn't get the right system directory when built from
anywhere other than waterfall root, because it was calculated at import
time.

### DIFF
--- a/src/compile_torture_tests.py
+++ b/src/compile_torture_tests.py
@@ -21,24 +21,6 @@ import sys
 
 import testing
 
-# TODO: Pass this in from build.py
-INSTALL_DIR = os.path.join(os.getcwd(),
-                           'src', 'work', 'wasm-install', 'sysroot')
-
-CFLAGS_COMMON = ['--std=gnu89', '-DSTACK_SIZE=1044480',
-                 '-w', '-Wno-implicit-function-declaration']
-CFLAGS_EXTRA = {
-    'wasm': ['--target=wasm32-unknown-unknown', '-S', '-O2',
-             '--sysroot=%s' % INSTALL_DIR],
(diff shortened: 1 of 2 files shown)


src/compile_torture_tests.py
```diff
@@ -21,24 +21,6 @@
 
 import testing
 
-# TODO: Pass this in from build.py
-INSTALL_DIR = os.path.join(os.getcwd(),
```
sbc100: Wouldn't a simpler fix be to simply replace this line with:

```
SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
INSTALL_DIR = os.path.join(SCRIPT_DIR,  'src', 'work', 'wasm-install', 'sysroot')
```

(i.e. its the bug just that getcwd() was using rather than SCRIPT_DIR?)

jgravelle-google: Probably yeah. It kind of accidentally evolved into the current form because I was taking the TODO too literally.
In theory it might be desirable to decouple the location of the sysroot path from this module, but 1) this still doesn't do that, and 2) there's no real pressure to do so right now.

Although, using `__file__` here means we add a relative path dependency to compile_torture_tests.py, so if we move this file things can break in non-obvious ways. Having considered that, I'd rather the dependency on file structure flow from build.py into this one, rather than them both re-deriving it, if that makes sense.

sbc100: OK.

How about passing INSTALL_SYSROOT directly from build.py?   When you can make --sysroot and option rather than hardcoding it on line 106?

jgravelle-google: Done, that's even better, thanks

sbc100:
lgtm

PR #76, opened 2016-09-19
Generate `latest` and `lkgr` files in `work`

sbc100:
Seems like a good idea to keep all generated files
in the same directory.  Without this this show
up as untracked files in the root.


jgravelle-google:
LGTM

PR #77, opened 2016-09-19
Rename 'emscripten' step to 'emscripten libs'

sbc100:
So that is is more meaningful.


jgravelle-google:
LGTM

dschuff:
It's more than just emscripten libs. It just so happens that 'installing' emscripten is really just copying some files, but that's the important bit.

PR #78, opened 2016-09-21
Make asmjs/wasm builds of emscripten libs independent

dschuff:
Currently if the wasm-backend build of the emscripten libs fails, the
asmjs build will not run. This PR makes them both run even if one fails.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -871,7 +871,7 @@ def Emscripten(use_asm=True):
           os.path.join(EMSCRIPTEN_SRC_DIR, 'tests', 'hello_libcxx.cpp'),
           '-O2', '-s', 'BINARYEN=1', '-s', 'BINARYEN_METHOD="native-wasm"'])
 
-    except:
+    except proc.CalledProcessError:
       # Note the failure but allow the build to continue.
       buildbot.Fail()
     finally:
--- a/src/build.py
+++ b/src/build.py
@@ -853,27 +853,30 @@ def Emscripten(use_asm=True):
                         EMSCRIPTEN_CONFIG_ASMJS)
   WriteEmscriptenConfig(os.path.join(SCRIPT_DIR, 'emscripten_config_vanilla'),
                         EMSCRIPTEN_CONFIG_WASM)
-  try:
-    # Build a C++ file with each active emscripten config. This causes system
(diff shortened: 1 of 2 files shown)


dschuff:
@kripken I think this fixes the errors I mentioned this morning. At least, it works locally.

src/build.py
```diff
@@ -853,27 +853,30 @@ def WriteEmscriptenConfig(infile, outfile):
...
           os.path.join(emscripten_dir, 'em++'),
           os.path.join(EMSCRIPTEN_SRC_DIR, 'tests', 'hello_libcxx.cpp'),
           '-O2', '-s', 'BINARYEN=1', '-s', 'BINARYEN_METHOD="native-wasm"'])
 
-  except proc.CalledProcessError:
-    # Note the failure but allow the build to continue.
-    buildbot.Fail()
-  finally:
-    del os.environ['EMCC_DEBUG']
+    except:
```
jgravelle-google: I don't think anything else in the try block can throw, but I'd think we're better off here leaving the explicit exception type, in case shenanigans?

dschuff: Ah, right I had taken that out as a local test and forgot to put it back in, good catch. so to speak.

jgravelle-google:
LGTM

src/build.py
```diff
@@ -853,27 +853,30 @@ def WriteEmscriptenConfig(infile, outfile):
...
-    # tests in parallel). Do it with full debug output.
-    # This depends on binaryen already being built and installed into the
-    # archive/install dir.
-    os.environ['EMCC_DEBUG'] = '2'
-    configs = [EMSCRIPTEN_CONFIG_WASM] + (
-        [EMSCRIPTEN_CONFIG_ASMJS] if use_asm else [])
-    for config in configs:
+
+  configs = [EMSCRIPTEN_CONFIG_WASM] + (
+      [EMSCRIPTEN_CONFIG_ASMJS] if use_asm else [])
```
sbc100: I find this needlessly cryptic. How about just:

```
configs = [EMSCRIPTEN_CONFIG_WASM]
if use_asm:
    configs.append(EMSCRIPTEN_CONFIG_ASMJS)
```

I know this change has already merged, but anyway...

PR #79, opened 2016-09-22
Don't copy 0xb wasm.js-post.js over master in the binaryen install

dschuff:
Also copy the rest of Binaryen's JS files


dschuff:
The conditional is pretty ugly, but this whole thing can go away as soon as v8 and wabbit are updated.

jgravelle-google:
LGTM

PR #80, opened 2016-09-23
Error out if passed non-existent/mis-spelled steps

sbc100:


src/build.py
```diff
@@ -611,9 +611,16 @@ def __init__(self, include=None, exclude=None):
...
   def Apply(self, targets):
     """ Return the filtered list of targets. """
+    all_names = [t.name for t in targets]
     if self.include is not None:
+      missing_includes = [i for i in self.include if i not in all_names]
+      if missing_includes:
+        raise Exception('Invalid step name(s): ' + str(missing_includes))
       return [t for t in targets if t.name in self.include]
     if self.exclude:
+      missing_excludes = [i for i in self.include if i not in all_names]
```
jgravelle-google: Should be `for i in self.exclude`

src/build.py
```diff
@@ -611,9 +611,16 @@ def __init__(self, include=None, exclude=None):
...
+      if missing_excludes:
+        raise Exception('Invalid step name(s): ' + str(missing_excludes))
```
jgravelle-google: Not too fond of the duplication, though my suggestion feels kinda icky as well.

Can hoist this out with something like
```python
specified_list = self.include or self.exclude or []
missing = # filter specified_list
if missing: # exception
```

Not sure if that's actually more readable, so leaving it might be fine.

dschuff: I like that pretty well. I might call it `specified_names` instead of `specified_list`.

dschuff:
This seems to have broken exclusions, e.g. `--sync-exclude=binaryen` or `--build-exclude=llvm`. Can you take a look?

sbc100:
Seems to work for me, and you be more specific.   I tried `./src/build.py --sync-exclude=llvm,clang,compiler-rt` and it worked as expected.   I cant' see how this change could have effect anything since all it does it raise exceptions.

dschuff:
Weird, any sync-exclude fails for me (it throws the exact exception that you mentioned)
```
src/build.py --sync-exclude=llvm,clang,compiler-rt
... [sync things]...
  File "src/build.py", line 1286, in main
  File "src/build.py", line 1162, in run
  File "src/build.py", line 661, in SyncRepos
  File "src/build.py", line 637, in Check
  File "src/build.py", line 624, in Apply
Exception: Invalid step name(s): ['llvm', 'compiler-rt']
```

jgravelle-google:
Same here, though I found it with `--test-include bare`

The problem is:
```python
  def Check(self, target):
    """ Return true if the specified target will be run. """
    return len(self.Apply([self.DummyTarget(target)])) > 0
```

So we call Apply on a 1-element list and that tends to not work

sbc100:
Oh crap.  Yes.  I will revert this an come up with a better fix.

jgravelle-google:
Well, https://github.com/WebAssembly/waterfall/issues/92 fixes it, so I'd just leave it

PR #81, opened 2016-09-23
Adding find svn rev script to avoid requiring git svn.

flagxor:
This may fix mac.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -61,6 +61,8 @@ BINARYEN_SRC_DIR = os.path.join(WORK_DIR, 'binaryen')
 BINARYEN_0xB_SRC_DIR = os.path.join(WORK_DIR, 'binaryen-0xb')
 MUSL_SRC_DIR = os.path.join(WORK_DIR, 'musl')
 
+FIND_SVN_REV = os.path.join(SCRIPT_DIR, 'find_svn_rev.py')
+
 PREBUILT_CLANG = os.path.join(WORK_DIR, 'chromium-clang')
 PREBUILT_CLANG_TOOLS_CLANG = os.path.join(PREBUILT_CLANG, 'tools', 'clang')
 PREBUILT_CLANG_BIN = os.path.join(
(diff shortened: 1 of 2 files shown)


src/build.py
```diff
@@ -526,15 +528,15 @@ def NoSync(*args):
 
 def CurrentSvnRev(path):
   return int(proc.check_output(
-      ['git', 'svn', 'find-rev', 'HEAD'], cwd=path).strip())
+      [FIND_SVN_REV, 'HEAD'], cwd=path).strip())
```
sbc100: Do these fix on one line now?

src/find_svn_rev.py
```diff
@@ -0,0 +1,70 @@
...
+    print m.group(1)
+    return 0
+  else:
+    sys.stderr.write('Commit not found!\n')
+    return 1
+
+
+def FindSvnRev(rev):
+  last_commit = None
+  p = subprocess.Popen(['git', 'log', 'HEAD'], stdout=subprocess.PIPE)
```
sbc100: Can this be done with `git log --grep`?

src/find_svn_rev.py
```diff
@@ -0,0 +1,70 @@
...
+import re
+import subprocess
+import sys
+
+
+COMMIT_MATCH = re.compile('^commit ([0-9a-f]+)')
+REV_MATCH = re.compile('git[-]svn[-]id[:].*[@]([0-9]+)')
+
+
+def FindGitCommit(commit):
```
dschuff: This function is confusingly named because it doesn't find/search for anything, and it returns an SVN rev not a git commit :)
How about `GetSvnRev` or `SvnRevFromGitCommit` or something like that?

src/find_svn_rev.py
```diff
@@ -0,0 +1,70 @@
...
+  p = subprocess.Popen(['git', 'log', 'HEAD'], stdout=subprocess.PIPE)
+  for line in p.stdout:
+    m = COMMIT_MATCH.search(line)
+    if m:
+      last_commit = m.group(1)
+    m = REV_MATCH.search(line)
+    if m:
+      if m.group(1) == rev:
+        if last_commit is None:
+          sys.stderr.write('Rev not found\n')
```
dschuff: this is rev found but no commit found?

src/build.py
flagxor: done

src/find_svn_rev.py
flagxor: done

src/find_svn_rev.py
flagxor: nixed

flagxor:
Ok, revised for both.

PR #84, opened 2016-09-24
Group archived assets under <buildername>/<buildnumber>

sbc100:
Now that we have multiple builders we should probably start
grouping assets rather than putting them all in single
directly.

For now, the build number is also included in the names
of the individual tar archives.

Also, drop the (now-redundant) `git/` and `svn/` distiction.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1268,10 +1268,10 @@ def main():
   test_filter = Filter(test_include, options.test_exclude)
 
   if IsBuildbot():
-    # Chrome's buildbot infra includes in its paths a module called 'tools' which
-    # conflicts with emscripten's own 'tools' module and overrides the emscripten
-    # test runner's import. We don't need that infra in this script, so we just
-    # scrub it from the environment.
+    # Chrome's buildbot infra includes in its paths a module called 'tools'
+    # which conflicts with emscripten's own 'tools' module and overrides the
+    # emscripten test runner's import. We don't need that infra in this script,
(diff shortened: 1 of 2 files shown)


sbc100:
@dschuff @jgravelle-google am I correct in thinking the svn/ and git/ folders don't serve any purpose?

src/build.py
```diff
@@ -280,12 +280,19 @@ def Tar(directory, print_content=False):
...
+  """Archive the file with the given name, and with the LLVM git hash."""
+  if not BUILDBOT_BUILDNUMBER:
+    return
+  buildbot.Link('download', cloud.Upload(local_name, '%s/%s/%s' % (
+      BUILDBOT_BUILDERNAME, BUILDBOT_BUILDNUMBER, remote_name)))
+
+
 def Archive(name, tar):
   """Archive the tar file with the given name, and with the LLVM git hash."""
-  if not BUILDBOT_BUILDERNAME:
```
dschuff: It looks like we have a few places in this file where we test on one or the other of `BUILDBOT_BUILDERNAME` or `BUILDBOT_BUILDNUMBER` to just early-return from a function that does buildbot stuff. Maybe it makes sense to make a global function `IsBuildbot()` and use that? And maybe it can assert that if we think we are a buildbot that we have all the correct info in the env.

dschuff:
flake8 is unhappy with line length but otherwise LGTM (also "asserts"->"assets" in the PR title).

PR #88, opened 2016-09-27
Set fetch depth as well as clone depth for torture test suite

dschuff:
Needed if we want to keep the shallow clone shallow.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -378,12 +378,14 @@ class Source:
     else:
       clone = ['git', 'clone', self.git_repo, self.src_dir]
       if self.depth:
-        clone.append('--depth')
-        clone.append(str(self.depth))
+        clone.extend(['--depth', str(self.depth)])
       proc.check_call(clone)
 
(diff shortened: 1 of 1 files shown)


dschuff:
It looks like the fetches are taking a lot longer for me now than previously; is anyone else seeing this? May be the case on the buildbot too. I reread the man page for `git fetch` and it seems to imply that it may not be necessary anyway if the clone was created as a shallow clone anyway.

sbc100:
I think you might be right... maybe a fetch with --depth=1 will actually end up downloading a lot more because it can't incrementally build on a previously clone/fetch.

jfbastien:
The GCC clone depth is 1000, and then we checkout a particular revision to avoid torture test cahnges breaking things. Is that revision more than 1000 revisions in the past?

dschuff:
No I don't think so; I just noticed that fetch was being slow when I
bootstrapped it on my laptop, and thought it was maybe because the fetch
was fetching more than was originally cloned.

PR #91, opened 2016-09-28
Consistent naming of asm2wasm and emwasm in steps

sbc100:


src/build.py
```diff
@@ -1243,12 +1243,12 @@ def run(sync_filter, build_filter, test_filter, options):
 
   if test_filter.Check('emwasm'):
     emscripten_wasm_out = CompileLLVMTortureBinaryen(
-        'Compile LLVM Torture (emscripten+wasm backend)',
+        'emwasm',
```
dschuff: The consistency is good, although I'm worried that it might not be obvious enough that "emwasm" is emscripten with the wasm backend. Before this patch we have the problem that it's not necessarily clear that the steps (emwasm and emscripten+wasm backend) corresponded to each other, and with this patch we have the problem that it's not obvious what emwasm is. We could have a step that has both names, or always use the longer name, but that might be a little verbose. We could use "vanilla" which is short and is what emscripten calls it, but I don't actually think that's a great name (I want to change it in emscripten :) ). So... I don't know. Ideas?

sbc100: I vote we go for consistency in the waterfall now, and try to come up with a better name than emwasm in a future CL.

dschuff: Sounds good to me.

PR #92, opened 2016-09-28
Make Filter.Check not throw on unknown targets

dschuff:
Throwing on unknown targets only makes sense when a list of possible
targets is given (e.g. when calling Apply), not when using bare
Check (since the filter does not necessarily contain all known targets
itself).

This refactors the actual exclusion logic into check, and implements
Apply in terms of that.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -611,7 +611,7 @@ class Filter:
     """
     if include and exclude:
       raise Exception('Filter cannot include both include and exclude rules')
-
+    print 'Filter: ', include, exclude
     self.include = include
     self.exclude = exclude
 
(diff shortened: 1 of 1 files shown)


dschuff:
Should fix the issues I saw with #80.

src/build.py
```diff
@@ -611,7 +611,7 @@ def __init__(self, include=None, exclude=None):
     """
     if include and exclude:
       raise Exception('Filter cannot include both include and exclude rules')
-
+    print 'Filter: ', include, exclude
```
jgravelle-google: Debugging print

src/build.py
```diff
@@ -622,19 +622,17 @@ def Apply(self, targets):
...
+
+    return [t for t in targets if self.Check(t.name)]
 
   def Check(self, target):
     """ Return true if the specified target will be run. """
-    return len(self.Apply([self.DummyTarget(target)])) > 0
+    if self.include is not None:
+      return target in self.include
+
+    return self.exclude is None or target not in self.exclude
```
jgravelle-google: This might be more readable as
```python
if self.exclude is not None:
  return target not in self.exclude
return True
```

src/build.py

src/build.py

src/build.py
```diff
@@ -623,16 +623,16 @@ def Apply(self, targets):
...
-
     return [t for t in targets if self.Check(t.name)]
 
   def Check(self, target):
     """ Return true if the specified target will be run. """
     if self.include is not None:
       return target in self.include
 
-    return self.exclude is None or target not in self.exclude
-
```
jgravelle-google: The flake8 didn't like the extra line here? Interesting I thought it liked two blank lines... I guess that's only for top-level definitions.

dschuff: yes, 2 lines for top level functions, 1 line for methods.

PR #93, opened 2016-09-28
Revert "Error out if passed non-existent/mis-spelled steps"

sbc100:
Reverts WebAssembly/waterfall#80

This caused issues whenever filter.Check() was called.


sbc100:
Fixed in #92

PR #96, opened 2016-09-28
Test running refactoring

jgravelle-google:
* Add a Test class to mirror Build and Sync steps, unify filtering logic
* Add tests to --help message
* Add more details to error messages when sync/build/testing a nonexistent step

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1201,9 +1201,11 @@ def ParseArgs():
                                            initial_indent='  ',
                                            subsequent_indent='  '))
 
-  epilog = (
-      TextWrapNameList('sync targets:\n', ALL_SOURCES) + '\n\n' +
-      TextWrapNameList('build targets:\n', AllBuilds()))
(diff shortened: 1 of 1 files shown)


sbc100:
Nice. I was working on the exact same change :)

BTW, I think you lost the cleanups I made in #91 somehow.

jgravelle-google:
I started working on it because I thought it was the way to implement #92. It doesn't quite cover all the cases to fix that, but it still seemed worthwhile to do.

And I think that's because I started my branch before you merged those? Or I was lazy about syncing. Your cleanups are still on master, so I think we're good.

sbc100:
I think some changes were lost somehow (this change re-introduced `emscripten+wasm backend` which I had removed).

jgravelle-google:
Ah, you're right

I'd moved the tests, and didn't think too hard about why that would cause a merge conflict. My bad!

PR #103, opened 2016-09-29
Add build_info.json to archived binary package

sbc100:
Also, upload this file as lkgr/latest rather than creating
seperate local lkgr and latest files.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1054,17 +1054,16 @@ def Summary(repos):
   for step in buildbot.FailedList():
     print '    %s' % step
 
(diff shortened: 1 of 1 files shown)


dschuff:
@jgravelle-google does the 'sync-lkgr' bit you added recently depend on the local lkgr file or does it fetch it from the server every time?

jgravelle-google:
It fetches from server every time. The idea being that you can sync that from nothing and immediately get a working toolchain.

dschuff:
Cool, lgtm

PR #106, opened 2016-09-29
Update known test failures for 0xc

jgravelle-google:


src/build.py
```diff
@@ -1103,6 +1103,7 @@ def TestBare():
       runner=os.path.join(INSTALL_BIN, 'wasm.opt'),
       indir=s2wasm_out,
       fails=SPEC_KNOWN_TORTURE_FAILURES,
+      warn_only=True,  # TODO all the spec tests fail with syntax error
```
dschuff: I'll file a bug on the spec for this, and we can add it to this PR

dschuff: https://github.com/WebAssembly/spec/issues/353

dschuff: Oh wait this is actually because the spec hasn't been merged to 0xc yet. Once it does, then that bug will be the relevant one.
Actually we should update this PR to pin the spec to the 0xc branch, since I don't know if anyone will merge it until Andreas gets back (I'm not sure if there are any outstanding issues left in ml-proto).

binji: Yeah, we could do it, but it's probably better to wait for him.

jgravelle-google: It still fails with the 0xc branch, but at least it's a more accurate error message?

So here I'm thinking, pin to 0xc, leave `warn_only=True`, link the spec bug in the TODO

dschuff: yeah, sgtm

rossberg-chromium: Heads-up: spec/binary-0xc is now merged into master, and I fixed the offending WebAssembly/spec#353.

dschuff:
Thanks, updated in https://github.com/WebAssembly/waterfall/pull/127

PR #107, opened 2016-09-29
Consitent naming of emscripten steps

sbc100:
emscripten is now built in two different steps.  This
change renames the first step from just `emscripten`
to `emscripten (emwasm)` for consistency with other
pairs of steps in the build.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -838,7 +838,7 @@ def Fastcomp():
 
 
 def Emscripten(use_asm=True):
-  buildbot.Step('emscripten')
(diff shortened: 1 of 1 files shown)


dschuff:
Maybe we should put the asm2wasm one first, since we run the asm2wasm tests first?

sbc100:
Its a little trick to do that because the asm2wasm build is optional (depends on the use_asm argument).  I'll give it a go.

dschuff:
ah, well if it makes things much more complex it's probably not worth it.

PR #108, opened 2016-09-29
Add flag to show git statuses

jgravelle-google:
Not sure if this is useful to anyone else, but here's a quick way to query the status of all the things.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -413,6 +413,13 @@ class Source:
         'remote': remote,
     }
 
+  def PrintGitStatus(self):
+    """"Print the current git status for the sync target."""
+    print '<<<<<<<<<< STATUS FOR', self.name, '>>>>>>>>>>'
+    if self.src_dir:
+      proc.check_call(['git', 'status'], cwd=self.src_dir)
+    print
+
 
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -1233,10 +1240,19 @@ def SplitComma(arg):
       '--test-exclude', dest='test_exclude', default='', type=SplitComma,
       help='Include only the comma-separated list of test targets')
 
+  parser.add_argument(
+      '--git-status', dest='git_status', default=False, action='store_true',
+      help='Show git status for each sync target')
```
dschuff: maybe add "and exit" or "only show git statuses" or something like that.

src/build.py
```diff
@@ -413,6 +413,13 @@ def pretty(fmt):
         'remote': remote,
     }
 
+  def PrintGitStatus(self):
+    """"Print the current git status for the sync target."""
+    print '<<<<<<<<<< STATUS FOR', self.name, '>>>>>>>>>>'
+    if self.src_dir:
```
sbc100: `os.path.exists(self.src_dir)` ?

jgravelle-google: Good catch. That way it'll do something sane if you're only partially synced.

PR #109, opened 2016-09-29
Skip similarity testing

jgravelle-google:
When >1,000 tests fail, calculating similarity on them takes ~3minutes on my machine.
If >50% of tests are failing, they probably have one overarching failure, and any arbitrary test case's output should be sufficient. Either that or we have bigger problems.

### DIFF
--- a/src/testing.py
+++ b/src/testing.py
@@ -203,7 +203,12 @@ def execute(tester, inputs, fails):
   for result in results:
     sys.stdout.write(str(result) + '\n\n')
   cutoff = 0.9
-  similar_expected_failures = similarity(expected_failures, cutoff)
+  too_many_failures = int(len(inputs) * 0.5)
+  if len(expected_failures) < too_many_failures:
+    similar_expected_failures = similarity(expected_failures, cutoff)
+  else:
+    print 'Too many expected failures to show similarity'
+    similar_expected_failures = []
   for s in similar_expected_failures:
     tests = ' '.join(s.tests)
     if s.average >= cutoff * 100.:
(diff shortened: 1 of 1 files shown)


src/testing.py
```diff
@@ -202,33 +202,44 @@ def execute(tester, inputs, fails):
...
-      sample = [f for f in unexpected_failures if f.test == s.tests[0]][0]
-      sys.stdout.write('Sample failure: %s\n' % sample)
-    else:
-      sys.stdout.write(('\nUngrouped unexpected failures, '
-                        'average %s%% similarity with stddev %s: '
-                        '%s\n') % (s.average, s.stddev, tests))
+  # Calculating similarity is pretty expensive. If more than half the tests are
+  # failing, it can take minutes, and most of them are probably failing for the
+  # same fundamental reason. Skip in that case.
+  too_many_failures = int(len(inputs) * 0.5)
```
sbc100: Just `len(inputs) / 2` would be simpler, no?

dschuff: Yeah probably. although I guess you'd still have to convert to int because of future division.

sbc100: future division?

jgravelle-google: Yes and no. `0.5` here is mirroring the `cutoff` variable above (which isn't as co-located as I'd like because of the comment), in that it's a percentage. The way I expect this to change is to tweak that percentage, say 30% gets deemed too much, we would change it to `0.3`. The problem here is the comment, framing it as "half" instead of "50%".
(Okay realistically the only time we're changing this is to remove it completely.)
Better would be to change this to `too_much_failure_percent = 0.5`, and then recalculate off `len(inputs)` later on. Much better communicates the intent.

And really it doesn't need to be int-casted, because we only use the number as a comparison, and it just adds line noise.

jgravelle-google: https://github.com/WebAssembly/waterfall/pull/110

jgravelle-google:
So for some reason I originally read that as "test only the first 1024 failing test cases against each other", but implemented my thing anyway because ???

Running a quick experiment, it turns out line 50 says "only check the first 1024 characters of each test's output for similarity", so it doesn't!

jfbastien:
@jgravelle-google ah right, I forgot what I'd implemented. Sorry, carry on! :)

src/testing.py
dschuff: Re: future division: https://www.python.org/dev/peps/pep-0238

sbc100: the future is coming!

PR #111, opened 2016-09-30
Create emcc/em++ wrappers in toolchain bin directory

sbc100:
These wrappers set/override EM_CONFIG meaning that the user's emscripten config
won't interfere with the toolchain. They also set EMCC_SKIP_SANITY_CHECK so that
emscripten won't try to write to the toolchain directory (which is not writable
in the packaged version).

Also, add /etc/alternatives for the core set of binaries in the debian package
so they are accessible via /usr/bin.

Also, fix the pyc cleanup command, and don't remove the third_party directory
(it is needed for closure-compiler.jar at least).

### DIFF
--- a/src/emcc_wrapper.sh
+++ b/src/emcc_wrapper.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+SCRIPT_FILE=$(readlink -f "$BASH_SOURCE")
+SCRIPT_DIR=$(cd $(dirname $SCRIPT_FILE) && pwd)
+
+export EMCC_SKIP_SANITY_CHECK=1
+export EM_CONFIG=$(dirname $SCRIPT_DIR)/emscripten_config
+
+exec $SCRIPT_DIR/emscripten/$(basename $0) $*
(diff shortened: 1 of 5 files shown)


src/emcc_wrapper.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+SCRIPT_FILE=$(readlink -f "$BASH_SOURCE")
+SCRIPT_DIR=$(cd $(dirname $SCRIPT_FILE) && pwd)
+
+export EMCC_SKIP_SANITY_CHECK=1
```
dschuff: What things does this sanity check do that need to be skipped?

sbc100: It writes a file called <emconfig>_sanity alongside the emconfig (which is a non-writable location in our case).

sbc100:
@flagxor .. we talked about this wrapper script method earlier.

src/emcc_wrapper.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+SCRIPT_FILE=$(readlink -f "$BASH_SOURCE")
```
flagxor: ${BASH_SOURCE} ?

sbc100: Done

PR #113, opened 2016-09-30
Upload lkgr/latest file to per-builder sub-directory

sbc100:
Otherwise the three different builders will race to
update these files.

Also, add .json file extension so it clear what type
of file these are.

Also, and missing newline to end of json file.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -657,7 +657,7 @@ def SyncRepos(filter, sync_lkgr=False):
       buildername = 'mac'
     elif sys.platform == 'win32':
       buildername = 'windows'
-    else
+    else:
       raise Exception('Unknown platform: %s' % sys.platform)
     lkgr_file = os.path.join(WORK_DIR, 'lkgr.json')
     cloud.Download('%s/lkgr.json' % buildername, lkgr_file)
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -650,8 +651,14 @@ def SyncRepos(filter, sync_lkgr=False):
 
   good_hashes = None
   if sync_lkgr:
-    lkgr_file = os.path.join(WORK_DIR, 'lkgr')
-    cloud.Download('lkgr', lkgr_file)
+    if sys.platform.startswith('linux'):
+      buildername = 'linux'
+    elif sys.platform == 'darwin':
+      buildername = 'mac'
+    elif sys.platform == 'win32':
```
jgravelle-google: I'd add `else: assert "unexpected sys.platform"` for an easier to parse error message if this ever doesn't work

src/build.py
```diff
@@ -1067,12 +1074,12 @@ def Summary(repos):
...
     f.write(info_json)
 
-  buildbot.Link('latest', cloud.Upload(info_file, 'latest'))
+  buildbot.Link('latest.json', cloud.Upload(info_file, 'latest.json'))
 
   if buildbot.Failed():
     buildbot.Fail()
   else:
-    buildbot.Link('lkgr', cloud.Upload(info_file, 'lkgr'))
+    buildbot.Link('lkgr.json', cloud.Upload(info_file, 'lkgr.json'))
```
jgravelle-google: Is there a change in cloud.py to add the buildername as well?

src/build.py

sbc100:
The links on the waterfall are auto-generated from the code (or are you talking about some other links?)

sbc100:
Oh.  I see you mean the link in the header.. yes that link is already broken since it has 'git' in that path which we are not using anymore.  The code for that is part of chromium infra.. I'll find it and update it.

src/build.py
dschuff: I suppose the endgame is that we reimplement or copy `pynacl/platform.py`

jfbastien:
What @jgravelle-google said: the goal of lkgr is that you can go to wasm-stat.us and download the last known good revision. With multiple builders maybe you need some aggregation from multiple platforms before you update? Having one per platform is weird.

dschuff:
Yeah, we probably shouldn't call it 'lkgr' if it's not green on all the platforms :) Although I can't think of a really good way to do that with the buildbot setup the way way have. One slightly-ugly way I can think of might be to have each platform update its own staging lkgr (i.e. the lkgr for that platform); and then additionally check the lkgrs of the other platforms, and if they are all up to the same revision, update the global lkgr. While that's technically still racy (I think it could in theory go backward) it should at least have the property that it shouldn't set an lkgr that isn't actually good on all 3 platforms.

jfbastien:
@dschuff that sounds good.

PR #116, opened 2016-10-04
Package emscripten in the root of the toolchain rather than under `bin`

sbc100:
The bin directory is normally reserved for executable that
should be in the PATH, not entire subdirectoris.


jgravelle-google:
I have no objection. @dschuff might have a better holistic sense of what belongs where, but this makes sense to me.

dschuff:
I think the original idea was that the emscripten bin directory is also... well, emscripten doesn't really have a bin directory; it's just the root of the emscripten source checkout, and there are a bunch of files in there that you wouldn't want polluting your bin directory, especially if you had it in your PATH. Maybe the thing to do is put a symlink or shell wrapper for `emcc` and `em++` instead.

sbc100:
I already created shell wrappers in #83

PR #121, opened 2016-10-07
Install fastcomp without static libs

dschuff:
The trunk LLVM install uses INSTALL_TOOLCHAIN_ONLY to avoid getting all
the static libraries, and copies a few extra tools. Use the same logic
for fastcomp, which should cut down the tarball size.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -239,21 +239,20 @@ def CopyTree(src, dst):
       shutil.copy2(os.path.join(root, f), dstfile)
 
 
-def CopyBinaryToArchive(binary, extra_dir=None):
+def CopyBinaryToArchive(binary, subdir=''):
   """All binaries are archived in the same tar file."""
-  install_bin = INSTALL_BIN
-  if extra_dir is not None:
-    install_bin = os.path.join(INSTALL_BIN, extra_dir)
+  install_bin = os.path.join(INSTALL_DIR, subdir, 'bin')
   print 'Copying binary %s to archive %s' % (binary, install_bin)
   Mkdir(install_bin)
   shutil.copy2(binary, install_bin)
 
 
-def CopyLibraryToArchive(library):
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -694,6 +693,23 @@ def Which(name):
     return ''
 
 
+def CopyLLVMTools(build_dir, install_subdir=''):
+  # The following isn't useful for now, and takes up space.
+  Remove(os.path.join(INSTALL_DIR, install_subdir, 'bin', 'clang-check'))
+  # The following are useful, LLVM_INSTALL_TOOLCHAIN_ONLY did away with them.
+  extra_bins = ['FileCheck', 'lli', 'llc', 'llvm-as', 'llvm-dis', 'llvm-link',
+                'llvm-nm', 'opt']
+  extra_libs = ['libLLVM*.so']
```
sbc100: I think you can remove this line now that I fixed it upstream to be installed with LLVM_INSTALL_TOOLCHAIN_ONLY

src/build.py
```diff
@@ -694,6 +693,23 @@ def Which(name):
     return ''
 
 
+def CopyLLVMTools(build_dir, install_subdir=''):
```
sbc100: I would use 'prefix=' here and above (or at least be consistent).  'prefix' is the autoconf name for this kind of thing.

dschuff: Yeah, except it's not really a prefix, it's more of an infix (it is actually used as a subdir of the `INSTALL_DIR` which is what you might call the global prefix).

sbc100: In autoconf terminology this is called the PREFIX, and the its appended to the DESTDIR.   So you might configure with PREFIX=/usr/local and then install with DESTDIR=/tmp/location, and you'd end up with all your files in /tmp/location/usr/local.

PR #123, opened 2016-10-10
Run with shell=True for git + gclient on windows.

flagxor:
PTAL

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -254,6 +254,18 @@ def CopyLibraryToArchive(library, install_subdir=''):
   shutil.copy2(library, install_lib)
 
 
+def Git(cmd, **kwargs):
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -255,6 +255,18 @@ def CopyLibraryToArchive(library, install_subdir=''):
   shutil.copy2(library, install_lib)
 
 
+def Git(cmd, **kwargs):
+  """Invoke git."""
+  return proc.check_output(
+      ['git'] + cmd, **kwargs, shell=sys.platform=='win32')
```
dschuff: Travis doesn't like this syntax. I think `**kwargs` has to go after all the manual keyword args?

flagxor:
PTAL

PR #124, opened 2016-10-11
Run python scripts under sys.executable

dschuff:
Because Windows.


jfbastien:
Why not add it directly in the `proc` module?

dschuff:
Apparently I raced with @flagxor on #125.
@jfbastien That's not a bad idea; I suppose it depends on how transparent a proxy to `suprocess` we want to keep `proc`. It might be really nice if `proc` could just automatically handle use of `sys.executable` for Python scripts and also adding `shell=True` as in #123 if necessary (but hopefully only when necessary). That might make for some fairly complex and magical logic but it would make the main script simpler...

sbc100:
Can this be closed now?

PR #125, opened 2016-10-11
Use sys.executable so python works on windows.

flagxor:


flagxor:
Submitting TBR.

jfbastien:
c.f. #124 I'm not sure that's the best fix.

PR #126, opened 2016-10-11
Use 'prefix' instead of 'install_subdir' in Copy routines

dschuff:


dschuff:
@sbc100

dschuff:
Followup to #121

PR #130, opened 2016-10-18
Add the emscripten testsuite ran with asm2wasm

jgravelle-google:
We're shipping asm2wasm as the officially supported toolchain for dev
preview, we'd like to make sure it's stable with V8.

- Moved ExecuteEmscriptenTestSuite closer to its usage
- asm2wasm testsuite should be an error if anything fails

This has a dependency on https://github.com/kripken/emscripten/pull/4633 , which disables the one currently known failing test.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1221,7 +1221,7 @@ def ExecuteEmscriptenTestSuite(name, config, outdir, warn_only):
 
 def TestEmtest():
   ExecuteEmscriptenTestSuite(
-      'emtest',
+      'emwasm',
       EMSCRIPTEN_CONFIG_WASM,
       EMSCRIPTEN_TEST_OUT_DIR,
       warn_only=True)
@@ -1229,7 +1229,7 @@ def TestEmtest():
 
 def TestEmtestAsm2Wasm():
   ExecuteEmscriptenTestSuite(
-      'emtest-asm2wasm',
+      'asm2wasm',
       EMSCRIPTEN_CONFIG_ASMJS,
       EMSCRIPTEN_ASMJS_TEST_OUT_DIR,
       warn_only=False)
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -1218,18 +1206,41 @@ def TestEmwasm():
...
+      EMSCRIPTEN_ASMJS_TEST_OUT_DIR,
+      warn_only=False)
 
 
 ALL_TESTS = [
     Test('bare', TestBare),
     Test('asm', TestAsm),
     Test('emwasm', TestEmwasm),
     Test('emtest', TestEmtest),
+    Test('emtest-asm2wasm', TestEmtestAsm2Wasm),
```
dschuff: The above means that "emtest" is only used as an abbreviation for the command line, which I think was the original intent? Now here we sort of have a sparse 2D matrix with test suite (torture, emscripten) x codegen (bare, asm2wasm, emwasm). `bare` is torture+bare, `asm` is torture+asm2wasm, `emwasm` is torture+emwasm, `emtest` is emscripten+emwasm, and now the new emscripten+asm2wasm. 

The intent was to keep these names short since we (at least I) type them on the command line semi-regularly; I don't know if that's more important than organization here or what. Just musing I guess, for now maybe this could be called `emtest-asm` to match `asm` above.

src/build.py
```diff
@@ -1218,18 +1206,41 @@ def TestEmwasm():
...
+         os.path.join(INSTALL_DIR, 'emscripten', 'tests', 'runner.py'),
+         'binaryen2', '--em-config', config],
+        cwd=outdir)
+  except proc.CalledProcessError:
+    buildbot.Fail(warn_only)
+
+
 def TestEmtest():
   ExecuteEmscriptenTestSuite(
       'emtest',
```
dschuff: This should probably be "emwasm" to match the way the other step names are formatted.

src/build.py
```diff
@@ -1218,18 +1206,41 @@ def TestEmwasm():
...
       EMSCRIPTEN_CONFIG_WASM,
-      EMSCRIPTEN_TEST_OUT_DIR)
+      EMSCRIPTEN_TEST_OUT_DIR,
+      warn_only=True)
+
+
+def TestEmtestAsm2Wasm():
+  ExecuteEmscriptenTestSuite(
+      'emtest-asm2wasm',
```
dschuff: Likewise this can be just "asm2wasm"

PR #132, opened 2016-10-20
Use CMake install for Binaryen instead of manual script

dschuff:
Depends on https://github.com/WebAssembly/binaryen/pull/798


sbc100:
This is way better!  Thanks.

PR #135, opened 2016-10-26
Make script output quieter when not on a buildbot

dschuff:
Don't print anything when not syncing anything
Don't collect and write/print repo info when not a buildbot


dschuff:
@jgravelle-google @sbc100

src/build.py
```diff
@@ -1318,7 +1329,10 @@ def run(sync_filter, build_filter, test_filter, options):
   Chdir(SCRIPT_DIR)
   Mkdir(WORK_DIR)
   SyncRepos(sync_filter, options.sync_lkgr)
-  repos = GetRepoInfo()
+  if IsBuildbot():
```
jgravelle-google: Cute idea: `repos = GetRepoInfo() if IsBuildbot() else {}`

PR #137, opened 2016-10-31
Disable llvm regression test run on Mac

dschuff:
It's broken with LLVM_LINK_LLVM_DYLIB until
https://reviews.llvm.org/D25304 lands.


dschuff:
Also add a few useful LLVM tools, in anticipation of https://reviews.llvm.org/D25590 adding wasm support to them.

PR #141, opened 2016-11-01
Use prebuilt binary node.js

dschuff:
Node.js isn't part of the base image for our buildbots, so it disappears whenever they get swapped or wiped. Use a prebuilt from nodejs.org (mirrored locally) instead. Keep the config file's fallback to /usr/bin/node in case someone ends up using that config file. Also use node as the compiler engine, as that more closely reflects what users are likely to do.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -122,7 +122,7 @@ OCAML_DIR = os.path.join(WORK_DIR, OCAML_VERSION)
 OCAML_OUT_DIR = os.path.join(WORK_DIR, 'ocaml-out')
 OCAML_BIN_DIR = os.path.join(OCAML_OUT_DIR, 'bin')
 
-# Use prebuilt Node.js binary because the buildbots don't have node preinstalled
+# Use prebuilt Node.js because the buildbots don't have node preinstalled
 NODE_VERSION = '7.0.0'
 NODE_BASE_NAME = 'node-v' + NODE_VERSION + '-'
 
--- a/src/build.py
+++ b/src/build.py
@@ -112,14 +112,18 @@ WATERFALL_REMOTE = '_waterfall'
 
 # Sync OCaml from a cached tar file because the upstream repository is only
 # http. The file untars into a directory of the same name as the tar file.
-OCAML_STORAGE_BASE = 'https://wasm.storage.googleapis.com/'
+WASM_STORAGE_BASE = 'https://wasm.storage.googleapis.com/'
 OCAML_VERSION = 'ocaml-4.02.2'
 OCAML_TAR_NAME = OCAML_VERSION + '.tar.gz'
 OCAML_TAR = os.path.join(WORK_DIR, OCAML_TAR_NAME)
-OCAML_URL = OCAML_STORAGE_BASE + OCAML_TAR_NAME
+OCAML_URL = WASM_STORAGE_BASE + OCAML_TAR_NAME
 OCAML_DIR = os.path.join(WORK_DIR, OCAML_VERSION)
 OCAML_OUT_DIR = os.path.join(WORK_DIR, 'ocaml-out')
 OCAML_BIN_DIR = os.path.join(OCAML_OUT_DIR, 'bin')
+
+# Sync Node.js binary
+NODE_VERSION = '7.0.0'
+NODE_TAR_BASE = 'node-v' + NODE_VERSION + '-'
 
 # Known failures.
 IT_IS_KNOWN = 'known_gcc_test_failures.txt'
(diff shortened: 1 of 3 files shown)


src/build.py
```diff
@@ -112,15 +112,25 @@
 
 # Sync OCaml from a cached tar file because the upstream repository is only
```
jgravelle-google: This comment is now slightly stale. Should probably mention node as a dependency for this url

src/emscripten_config_vanilla
```diff
@@ -20,7 +20,11 @@ BINARYEN_ROOT = os.path.join(WASM_INSTALL) # directory
 # EMSCRIPTEN_NATIVE_OPTIMIZER='/path/to/custom/optimizer(.exe)'
 
 # See below for notes on which JS engine(s) you need
-NODE_JS = os.path.expanduser(os.getenv('NODE') or '/usr/bin/nodejs') # executable
+prebuilt_node = '{{PREBUILT_NODE}}'
```
jgravelle-google: Duplicated changes are :(

Does it make sense to merge these into `emscripten_config_template`, and add another rewrite step for `{{LLVM_BIN_DIR}}` or so? That seems to be the only thing different between these files

src/build.py
```diff
@@ -875,6 +901,7 @@ def Emscripten(use_asm=True):
   def WriteEmscriptenConfig(infile, outfile):
     with open(infile) as config:
       text = config.read().replace('{{WASM_INSTALL}}', INSTALL_DIR)
```
jgravelle-google: Not sure if it's reasonable to chain these like
```python
text = (config.read()
    .replace('{{WASM_INSTALL}}', INSTALL_DIR)
    .replace('{{PREBUILD_NODE}}', NODE_BIN)
    )
```

I prefer that because there's less line noise but it's pretty much equivalent. Maybe if we add another thing here?

src/build.py
```diff
@@ -488,18 +498,32 @@ def SyncPrebuiltCMake(name, src_dir, git_repo):
...
-    proc.check_call(['tar', '-xvf', OCAML_TAR], cwd=WORK_DIR)
-    assert os.path.isdir(OCAML_DIR), 'Untar should produce %s' % OCAML_DIR
+  return SyncTarball(src_dir, 'OCaml', OCAML_VERSION, OCAML_URL, OCAML_TAR)
+
+
+def SyncPrebuiltNodeJS(name, src_dir, git_repo):
+  extension = {'darwin': 'gz', 'linux2': 'xz'}[sys.platform]
+  out_dir = os.path.join(WORK_DIR, NODE_BASE_NAME + NodePlatformName())
+  tarball = NODE_BASE_NAME + NodePlatformName() + '.tar.' + extension
+  print tarball
```
jgravelle-google: Debug print? Might be useful with more context than just a filename

src/build.py
```diff
@@ -488,18 +498,32 @@ def SyncPrebuiltCMake(name, src_dir, git_repo):
...
-    print 'Info: %s' % f.info()
-    with open(OCAML_TAR, 'wb') as out:
-      out.write(f.read())
-    proc.check_call(['tar', '-xvf', OCAML_TAR], cwd=WORK_DIR)
-    assert os.path.isdir(OCAML_DIR), 'Untar should produce %s' % OCAML_DIR
+  return SyncTarball(src_dir, 'OCaml', OCAML_VERSION, OCAML_URL, OCAML_TAR)
+
+
+def SyncPrebuiltNodeJS(name, src_dir, git_repo):
+  extension = {'darwin': 'gz', 'linux2': 'xz'}[sys.platform]
```
jgravelle-google: Should we make a struct with the platform specific junk in it? So here we could call `PlatformJunk().zipExtension` and elsewhere have

```python
RAW_JUNK = {
  'darwin': {
    zipExtension: 'gz',
    nodePlatformName: 'darwin-x64',
# ...
}
def PlatformJunk():
  return RAW_JUNK[sys.platform]
```

src/build.py
dschuff: OCaml hasn't changed, and the comment is still correct. But maybe we could be more clear what it applies to by moving `WASM_STORAGE_BASE` before it or something.

src/build.py
dschuff: Just a debug print, removed.

src/build.py
dschuff: I dunno, your version looks like more line noise to me... :) (and more lines :D)

src/emscripten_config_vanilla
dschuff: Yeah maybe. But probably a separate change.

src/build.py
dschuff: Maybe. Right now there's just one use. Once we start in on Windows, I expect what abstractions we actually need will become much more clear, so I'm inclined to wait until then.
Eventually we are just going to reinvent https://chromium.googlesource.com/native_client/src/native_client/+/master/pynacl/platform.py

src/build.py
dschuff: (but yeah if we add another I'd be inclined to do it that way)

src/build.py
```diff
@@ -110,9 +110,10 @@
 # clobbering any local development.
 WATERFALL_REMOTE = '_waterfall'
 
+WASM_STORAGE_BASE = 'https://wasm.storage.googleapis.com/'
+
 # Sync OCaml from a cached tar file because the upstream repository is only
 # http. The file untars into a directory of the same name as the tar file.
-WASM_STORAGE_BASE = 'https://wasm.storage.googleapis.com/'
```
jgravelle-google: Right, missed the dozen other OCaml-specific things underneath

src/build.py
jgravelle-google: Seems like a good idea. Should be simple enough to grep for `sys.platform` to see what cares as well.

src/emscripten_config_vanilla
jgravelle-google: Almost-definitely a separate change

PR #143, opened 2016-11-02
Fold fixes for python scripts and windows executables into proc

dschuff:
Instead of using sys.executable and windows extensions explicitly, make proc handle them automatically.

### DIFF
--- a/src/proc.py
+++ b/src/proc.py
@@ -31,7 +31,7 @@ from subprocess import * # flake8: noqa
 
 def FixPythonAndWindows(cmd):
   if cmd[0].endswith('.py'):
-    return [sys.executable].extend(cmd)
+    return [sys.executable] + cmd
 
   exe = cmd[0]
   def IsExe(f):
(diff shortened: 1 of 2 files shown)


dschuff:
Related to @jfbastien's comments in #124. I'm a little ambivalent about this because it's pretty magical for something that's currently just a transparent proxy to subprocess. But it is handy.

src/proc.py
```diff
@@ -29,9 +29,24 @@
...
+  if cmd[0].endswith('.py'):
+    return [sys.executable].extend(cmd)
+
+  exe = cmd[0]
+  def IsExe(f):
+    return os.path.isfile(f) and os.access(f, os.X_OK)
+  if not IsExe(exe):
+    if IsExe(exe + '.exe'):
+      exe = exe + '.exe'
+    if IsExe(exe + '.bat'):
```
jgravelle-google: Maybe if/elif/else throw-with-decent-error-message?

src/proc.py
```diff
@@ -29,9 +29,24 @@
 from subprocess import * # flake8: noqa
 
 
+def FixPythonAndWindows(cmd):
+  if cmd[0].endswith('.py'):
+    return [sys.executable] + cmd
+
+  exe = cmd[0]
+  def IsExe(f):
```
dschuff: Actually this isn't going to work unless we also essentially implement `which` by walking through PATH, or using `shell=True` which would mean we wouldn't need to use the file extensions anyway. Neither of those seems very appealing to me.

jgravelle-google:
Oh nothing was actually using the exe magic yet? Yeah probably for the best to just remove for now then.
Lgtm

dschuff:
Yeah, well I had forgotten to remove `Git()` and `GClient` from `build.py`

PR #144, opened 2016-11-03
Make sure the we only append "python" to python scripts that aren't i…

jgravelle-google:
…n PATH

gsutil.py is in PATH, so `python gsutil.py` doesn't work.
This won't fix Windows, but should fix the Mac and Linux bots.

### DIFF
--- a/src/proc.py
+++ b/src/proc.py
@@ -30,7 +30,8 @@ from subprocess import * # flake8: noqa
 
 
 def FixPython(cmd):
-  if cmd[0].endswith('.py'):
+  script = cmd[0]
+  if script.endswith('.py') and os.path.exists(script):
     return [sys.executable] + cmd
   return cmd
 


src/proc.py
```diff
@@ -30,7 +30,8 @@
 
 
 def FixPython(cmd):
-  if cmd[0].endswith('.py'):
+  script = cmd[0]
+  if script.endswith('.py') and os.path.exists(script):
```
dschuff: This could use the same logic as `IsExe()` in the previous version of the patch that broke all of this

dschuff:
By the way, it's totally fine (and often better) to just revert patches that break things, rather than add fixups. Especially if fixups even remotely nontrivial (or should i say, if they are not super trivial). There often ends up being more fixups when fixups are rushed, and auditing is easier that way too.

jgravelle-google:
Yeah I forgot that reverting other peoples' code was an option. That would've been more prudent.
In my head this was trivial enough that I felt confident I had it fully tested, and that I'd be watching the bots to make sure it worked.

src/proc.py
jgravelle-google: *Ah*, yep. Forgot `.access` existed between reviewing that and now.

PR #145, opened 2016-11-03
Implement 'which' equivalent

jgravelle-google:


jgravelle-google:
Second commit added after noticing that #144 successfully broke Windows, but not how I was expecting it to.
https://wasm-stat.us/builders/windows/builds/1569/steps/V8/logs/stdio

Linux/Mac are currently also failing to prepend '.py' when we add a `cwd=`, but they know how to execute a .py directly and don't seem to care.

src/proc.py
```diff
@@ -29,16 +29,27 @@
 from subprocess import * # flake8: noqa
 
 
-def FixPython(cmd):
+def Which(filename, cwd):
+  to_search = ([cwd] + os.environ['PATH'].split(os.pathsep))
```
dschuff: does the RHS need parens around it here?

src/proc.py
```diff
@@ -29,16 +29,27 @@
...
+  for path in to_search:
+    abs_path = os.path.join(path, filename)
+    if os.path.exists(abs_path):
```
dschuff: Should use the same logic as `IsExe` from https://github.com/WebAssembly/waterfall/pull/143/commits/5d9cc2e32e2242c6c5c08c6051e4f6b7c1e20cd3

src/proc.py
```diff
@@ -29,16 +29,27 @@
...
+      return abs_path
+  raise Exception('File "%s" not found. (cwd=`%s`, PATH=`%s`'
```
dschuff: could raise a `CalledProcessError` like what `subprocess` does in this case.

src/proc.py
jgravelle-google: Oh, no. It did at some point, but now it does not.

src/proc.py
jgravelle-google: [Huh, TIL.](http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html#tag_08_03)

Also TIL, [bash returns the first file it sees in your PATH if none are executable](http://git.savannah.gnu.org/cgit/bash.git/tree/findcmd.c#n553), presumably to give you a better error message? `which` is more directly standards-compliant in that it doesn't return anything when you ask `which non-executable-file-in-path`.

src/proc.py
jgravelle-google: Apparently that has a non-string constructor: https://fossies.org/dox/Python-3.5.2/subprocess_8py_source.html#l00380

Looking at that file, I'd like to use `SubprocessError`, which is closer, but python on my machine doesn't believe that it's defined.

src/proc.py
```diff
@@ -29,16 +29,31 @@
...
+      return abs_path
+  raise Exception('File "%s" not found. (cwd=`%s`, PATH=`%s`' %
+                  (filename, cwd, os.environ['PATH']))
+
+
+def FixPython(cmd, cwd):
   script = cmd[0]
-  if script.endswith('.py') and os.path.exists(script):
-    return [sys.executable] + cmd
+  if script.endswith('.py') and not CanExecute(script):
```
dschuff: I don't think we need to require that python scripts be executable (if you run with `python /path/to/foo.py` it doesn't need to be for it to work). I guess this is sort of the hazard of mixing python handling with executable handling. For executables it's pretty clear what we should do. 
For python, we should at least allow a non-executable script to run if the path given is an absolute path or is a file that exists relative to the cwd. Otherwise I guess we could search the path via Which for the script (including the requirement that it's executable).

PR #146, opened 2016-11-07
Update asm2wasm known failures

dschuff:
These were fixed in https://github.com/WebAssembly/binaryen/pull/821

### DIFF
--- a/src/test/asm2wasm_run_known_gcc_test_failures.txt
+++ b/src/test/asm2wasm_run_known_gcc_test_failures.txt
@@ -17,9 +17,7 @@ alloca-1.c.js
 bitfld-3.c.js
 bitfld-5.c.js
 builtin-bitops-1.c.js
-conversion.c.js
 eeprof-1.c.js
-floatunsisf-1.c.js
 frame-address.c.js
 pr17377.c.js
 pr32244-1.c.js


dschuff:
@kripken so we had this bug all along, and fixing it caused known failures to start working (i.e. the test caught the failure, we just didn't bother to triage all the tests). We have some more known failures (https://github.com/WebAssembly/waterfall/blob/master/src/test/asm2wasm_run_known_gcc_test_failures.txt), maybe we should take a closer look at them. That particular file doesn't have any triage notes, but the other lists of known failures do, so some of those reasons for being broken (or reasons that we don't care) would apply to asm2wasm too.

kripken:
Yeah, we should take a look. How can I do that - where can I find those source files?

dschuff:
The waterfall is currently pinned to https://chromium.googlesource.com/chromiumos/third_party/gcc/+/b6125c702850488ac3bfb1079ae5c9db89989406

dschuff:
Specifically the https://chromium.googlesource.com/chromiumos/third_party/gcc/+/b6125c702850488ac3bfb1079ae5c9db89989406/gcc/testsuite/gcc.c-torture/execute/ directory

dschuff:
(you should just be able to build the C file to a js and run it in your JS shell of choice)

kripken:
Great, thanks (searching for those files is impossible, it finds lots of lists of exceptions like ours, but not the contents ;) ).

PR #147, opened 2016-11-08
Triage asm2wasm known failures

kripken:
As suggested by @dschuff in #146.

Sadly (?) nothing interesting turned up, was hoping for another interesting bug like that float thing. But I guess it's good to know that nothing that should work is broken. In particular, no failure is wasm-specific, it all is stuff expected to fail in asm.js too:

 * The `returnaddress` and `frameaddress` builtins cannot be supported in asm.js or wasm.
 * Inline assembly stuff. (This is maybe the one thing we could make work, we could add some workaround to ignore them I guess, they seem like memory barriers, but that's not something we need to test yet anyhow.)
 * Stuff that aborts in native clang - things that only work in gcc, weird builtins and such, or undefined behavior that clang handles differently than gcc (e.g. the first in that category is a float to int rounding thing).
 * SIMD stuff (some might work in asm.js, actually, but not really supported).
 * One test uses `signal()`.

### DIFF
--- a/src/test/asm2wasm_run_known_gcc_test_failures.txt
+++ b/src/test/asm2wasm_run_known_gcc_test_failures.txt
@@ -1,40 +1,49 @@
 # Expected failures from running torture tests from emcc/binaryen with asm2wasm
 # and binaryen interpret-binary mode
 
-20010122-1.c.js
-20030222-1.c.js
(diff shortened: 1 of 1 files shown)


dschuff:
LGTM, thanks for checking it out.

PR #148, opened 2016-11-16
Unify CMake binary sync and support Windows

dschuff:
* Use the official CMake binary distribution for all platforms, including Windows
* Unify CMake download with SyncTarball, which does the same thing
* Enable WABT and Binaryen builds on Windows

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -68,11 +68,7 @@ PREBUILT_CLANG_BIN = os.path.join(
 CC = os.path.join(PREBUILT_CLANG_BIN, 'clang')
 CXX = os.path.join(PREBUILT_CLANG_BIN, 'clang++')
 
-# The archive itself contains the 'cmake343' directory.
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -492,61 +495,66 @@ def SyncPrebuiltClang(name, src_dir, git_repo):
...
+  except urllib2.URLError as e:
+    print 'Error downloading %s: %s' % (url, e)
+    raise
+
+
+def SyncPrebuiltCMake(name, src_dir, git_repo):
+  if os.path.isdir(PREBUILT_CMAKE_DIR):
+    print 'Prebuilt CMake directory already exists'
+  else:
+    platform = {'linux2': 'Linux',
```
jgravelle-google: nit: Apparently some systems are `'linux3'`, and Python3.3 and on have `sys.platform` just return `'linux'`, so the most-correct check here is `sys.platform.startswith('linux')` which mucks up the convenience of using a dict somewhat.

This is probably fine until we abstract out all this platform-y stuff.

PR #149, opened 2016-11-18
Fix debian package wrapper scripts

sbc100:
Tested with BananaBread.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e
 
-for x in emcc em++ wasm2wast wast2wasm wasm-interp s2wasm wasm-opt wasm-as \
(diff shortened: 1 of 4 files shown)


debian/postinst
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e
 
-for x in emcc em++ wasm2wast wast2wasm wasm-interp s2wasm wasm-opt wasm-as \
-         wasm-dis wasm-shell wasm.opt d8; do
+for x in emcc em++ emconfigure emmake wasm2wast wast2wasm wasm-interp s2wasm \
```
dschuff: how about wasmdump too?

sbc100: Done

PR #150, opened 2016-11-21
Support building CMake projects on Windows with Chromium infrastructure

dschuff:
Chrome's buildbots are special in that they do not have toolchains installed on the system. Instead they use a toolchain packaged and managed by depot_tools. This PR adds support for that setup.

* Factor toolchain stuff into host_toolchains.py
* Use scripts from Chromium's build tools repo to locate and use depot_tools toolchain
* Inject an environment like what CMake expects (similar to vcvars.bat) into build steps.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -485,6 +485,7 @@ def SyncToolchain(name, src_dir, git_repo):
   else:
     host_toolchains.SyncPrebuiltClang(name, src_dir, git_repo)
 
+
 def SyncArchive(out_dir, name, version, url):
   if os.path.isdir(out_dir):
     print '%s directory already exists' % name
@@ -576,7 +577,8 @@ ALL_SOURCES = [
     Source('host-toolchain', PREBUILT_CLANG,
            GIT_MIRROR_BASE + 'chromium/src/tools/clang',
            custom_sync=SyncToolchain),
-    Source('cr-buildtools', os.path.join(WORK_DIR, 'build'), GIT_MIRROR_BASE+'chromium/src/build'),
+    Source('cr-buildtools', os.path.join(WORK_DIR, 'build'),
+           GIT_MIRROR_BASE + 'chromium/src/build'),
     Source('cmake', '', '',  # The source and git args are ignored.
            custom_sync=SyncPrebuiltCMake),
     Source('nodejs', '', '',  # The source and git args are ignored.
--- a/src/build.py
+++ b/src/build.py
@@ -576,6 +576,7 @@ ALL_SOURCES = [
     Source('host-toolchain', PREBUILT_CLANG,
            GIT_MIRROR_BASE + 'chromium/src/tools/clang',
            custom_sync=SyncToolchain),
+    Source('cr-buildtools', os.path.join(WORK_DIR, 'build'), GIT_MIRROR_BASE+'chromium/src/build'),
     Source('cmake', '', '',  # The source and git args are ignored.
            custom_sync=SyncPrebuiltCMake),
     Source('nodejs', '', '',  # The source and git args are ignored.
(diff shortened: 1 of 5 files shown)


dschuff:
@flagxor

src/gyp.py
```diff
@@ -0,0 +1,8 @@
+# Dummy gyp.py because chromium/src/vs_toolchain.py depends on gyp, but it's in
```
dschuff: The ugliest bit is that the vs_toolchain.py script imports from some gyp-related code that lives in a different repo. We don't actually use that, so we stub it out. Another option is to use the version of that script in v8 (`v8/gypfiles/vs_toolchain.py`) which duplicates that functionality but doesn't depend on gyp. In which case we wouldn't need the fake gyp, but all the CMake builds would depend on the v8 build.

dschuff: Never mind; I decided to go ahead and use v8's script, to ensure that the same compiler version is used for v8 and the other builds (turns out v8 isn't quite in sync with Chromium).

src/host_toolchains.py
```diff
@@ -0,0 +1,87 @@
+#! /usr/bin/env python
+# -*- coding: utf-8 -*-
+
+#   Copyright 2015 WebAssembly Community Group participants
```
flagxor: 2016

src/build.py
```diff
@@ -838,12 +831,20 @@ def V8():
 def Wabt():
   buildbot.Step('WABT')
   Mkdir(WABT_OUT_DIR)
+  cc_env = None
+  if IsWindows():
+    cc_env = host_toolchains.SetUpEnv(WABT_OUT_DIR)
+    host_toolchains.CopyDlls(WABT_OUT_DIR, 'Release')
+    # TODO(dschuff): Figure out how to make thise statically linked?
```
binji: I think you just made up a word :)

src/host_toolchains.py
```diff
@@ -0,0 +1,87 @@
...
+  return ('chromium-clang', tools_clang)
+
+
+def SyncWinToolchain():
+  """Update the VS toolchain used by Chromium bots"""
+  proc.check_call([VS_TOOLCHAIN, 'update'])
+
+
+def GetEnv(dir):
+  """ Return the configured environment block as a python dict."""
```
binji: It's not too hard to guess the format of the file from the code below, but it might be nice to have a quick comment describing it.

src/host_toolchains.py
binji: 2016

src/host_toolchains.py
```diff
@@ -0,0 +1,87 @@
...
+    entries = f.read().split('\0')
+    for e in entries:
+      if not e:
+        continue
+      var, val = e.split('=', 1)
+      env[var] = val
+  return env
+
+
+def SetUpEnv(outdir):
```
binji: Seems like this is Windows-specific? Probably should be clearer in the function name.

src/build.py
dschuff: Done

src/host_toolchains.py
dschuff: Done.

src/host_toolchains.py
dschuff: Done

dschuff:
OK, I think I finally got everything.

src/build.py
```diff
@@ -33,6 +32,8 @@
 import cloud
 import compile_torture_tests
 import execute_files
+from file_util import Chdir, CopyTree, Mkdir, Remove
```
binji: Fine with me, I find this style to be pretty convenient.

PR #151, opened 2016-11-22
Fix Binaryen build on Windows

dschuff:


dschuff:
@flagxor might want to refactor to do the env setup part only once per build (rather than for each build target) but this should be enough to get the bot green.

flagxor:
lgtm

PR #153, opened 2016-12-01
Use tar command rather than python's tarfile to extract xz files

sbc100:


dschuff:
That's odd, was this broken before? The prebuilt node.js is xz already isn't it?

sbc100:
Yeah.. looks like the this is has been broken for a while, but the waterfall stayed green until I just did a clobber build.

sbc100:
I think you broke it in #148 :)

dschuff:
Yep, sure looks like it.

PR #154, opened 2016-12-08
Fix SyncPrebuiltClang to actually pull new versions of clang

dschuff:
Turns out we've been fetching but not actually updating the tools/clang
directory, so we haven't been getting the clang rolls.


dschuff:
I only noticed this because I tried to use goma locally and found that they finally dropped the old version of clang (from January!) from the server.

PR #156, opened 2016-12-13
Build LLVM on windows

dschuff:
LLVM's tests require unix utilities from GNUWin32, so download a prebuilt
of that, and inject into the PATH when building and testing LLVM.
Also disable building the LLVM dylib, as that seems broken on windows
currently.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -496,7 +496,6 @@ def SyncGNUWin32(name, src_dir, git_repo):
     return
   zipfile = os.path.join(GNUWIN32_DIR, GNUWIN32_ZIP)
   url = WASM_STORAGE_BASE + GNUWIN32_ZIP
-  print url
   return SyncArchive(GNUWIN32_DIR, name, '1', url)
 
 def NoSync(*args):
--- a/src/build.py
+++ b/src/build.py
@@ -725,12 +725,12 @@ def CopyLLVMTools(build_dir, prefix=''):
 
 
 def BuildEnv(build_dir, use_gnuwin32=False, bin_subdir=False,
-                runtime='Release'):
+             runtime='Release'):
   if not IsWindows():
     return None
   cc_env = host_toolchains.SetUpVSEnv(build_dir)
   if use_gnuwin32:
-    cc_env['PATH'] = cc_env['PATH'] + os.pathsep + os.path.join(GNUWIN32_DIR, 
+    cc_env['PATH'] = cc_env['PATH'] + os.pathsep + os.path.join(GNUWIN32_DIR,
                                                                 'bin')
   bin_dir = build_dir if not bin_subdir else os.path.join(build_dir, 'bin')
   Mkdir(bin_dir)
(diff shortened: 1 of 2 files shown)


src/build.py
```diff
@@ -482,6 +490,15 @@ def SyncPrebuiltNodeJS(name, src_dir, git_repo):
   return SyncArchive(out_dir, name, NODE_VERSION, node_url)
 
 
+# Utilities needed for running LLVM regression tests on Windows
+def SyncGNUWin32(name, src_dir, git_repo):
+  if not IsWindows():
+    return
+  zipfile = os.path.join(GNUWIN32_DIR, GNUWIN32_ZIP)
+  url = WASM_STORAGE_BASE + GNUWIN32_ZIP
+  print url
```
jfbastien: print?

src/build.py
```diff
@@ -705,9 +724,26 @@ def CopyLLVMTools(build_dir, prefix=''):
       CopyLibraryToArchive(os.path.join(build_dir, 'lib', e), prefix)
 
 
+def WinBuildEnv(build_dir, use_gnuwin32=False, bin_subdir=False,
+                runtime='Release'):
+  if not IsWindows():
+    return None
+  cc_env = host_toolchains.SetUpVSEnv(build_dir)
+  if use_gnuwin32:
+    cc_env['PATH'] = cc_env['PATH'] + os.pathsep + os.path.join(GNUWIN32_DIR,
```
jgravelle-google: Is there a subtle difference between `a + os.pathsep + os.path.join(b, c)` and `os.path.join(a, b, c)` ?

src/build.py
```diff
@@ -705,9 +724,26 @@ def CopyLLVMTools(build_dir, prefix=''):
...
+  Mkdir(bin_dir)
+  assert runtime in ['Release', 'Debug']
+  host_toolchains.CopyDlls(bin_dir, runtime)
+  return cc_env
+
+
 def LLVM():
   buildbot.Step('LLVM')
   Mkdir(LLVM_OUT_DIR)
+  cc_env = WinBuildEnv(LLVM_OUT_DIR, use_gnuwin32=True, bin_subdir=True)
```
jgravelle-google: This feels weird, because without knowing the implementation of `WinBuildEnv` this looks like `LLVM` assumes it's Windows.

Does it make sense to rename `WinBuildEnv` to `BuildEnv`? Then in theory if we have any non-Win env modifications they go there, and at the callsite it looks more consistent.

src/build.py
dschuff: `os.pathsep` is `;` (on windows), i.e. the character that separates the directories in `%PATH%` (not to be confused with `os.sep` which is the slash :D )

src/build.py
dschuff: Removed.

src/build.py
dschuff: Renamed to `BuildEnv`

PR #157, opened 2016-12-14
Disable build of libLTO in LLVM

dschuff:
Our mac bot has an old XCode with an ld64 that is too old for libLTO
as built on trunk, and it's causing the tests to fail.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -755,6 +755,8 @@ def LLVM():
              '-DCOMPILER_RT_INCLUDE_TESTS=OFF',
              '-DLLVM_BUILD_LLVM_DYLIB=%s' % build_dylib,
              '-DLLVM_LINK_LLVM_DYLIB=%s' % build_dylib,
+             # Our mac bot's toolchain's ld64 is too old for trunk libLTO.
+             '-DLLVM_TOOL_LTO_BUILD=OFF',
              '-DLLVM_INSTALL_TOOLCHAIN_ONLY=ON',
              '-DLLVM_ENABLE_ASSERTIONS=ON',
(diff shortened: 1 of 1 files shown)


jfbastien:
Is there a plan to update the Mac bots?

src/build.py
```diff
@@ -755,6 +755,8 @@ def LLVM():
              '-DCOMPILER_RT_INCLUDE_TESTS=OFF',
              '-DLLVM_BUILD_LLVM_DYLIB=%s' % build_dylib,
              '-DLLVM_LINK_LLVM_DYLIB=%s' % build_dylib,
+             # Our mac bot's toolchain's ld64 is too old for trunk libLTO.
+             '-DLLVM_TOOL_LTO_BUILD=OFF',
```
jgravelle-google: Assuming this is fine to disable on all platforms because we don't use LLVM's LTO anyway?

dschuff:
I don't plan to update them myself, because the installation is handled by chrome infra and I don't want to have to redo it whenever the slave gets replaced, etc. There is also a hermetic mac toolchain (similar to the windows one I set up for the windows bot) which I got working locally, but it's the same version as the installed one. However we are still using the chromium clang to compile, so it's really just the linker and binutils, etc that are old. So other than this it doesn't seem to be a problem.

src/build.py
jfbastien: I'd file a bug to re-enable though. We'll want to test LTO in the future, and allow folks to use this toolchain with LTO enabled.

dschuff: Right. libLTO (not to be confused with the LLVM gold plugin, which also does LTO) is really only used for mac's ld64 AFAIK, even though you can also build it on Linux. And since we aren't using this libLTO for wasm, it's find to disable for now.

PR #160, opened 2016-12-16
Use git bash from depot_tools for LLVM regression tests

dschuff:
Several of clang's regression tests use some nontrivial unix utilities
and/or bash features. We've been using the gnuwin32 ports of these
utilities, but for reasons I don't fully understand yet, this
reliably causes some issues with temp files created by tests on
windows 7 (windows 10 seems fine). The git that comes with
depot_tools has an msys bash with msys versions of these utilities
that are sufficient for running the LLVM tests, so switch to that
for now.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -767,12 +767,14 @@ def LLVM():
 
   proc.check_call(command, cwd=LLVM_OUT_DIR, env=cc_env)
   proc.check_call(['ninja', '-v'] + jobs, cwd=LLVM_OUT_DIR, env=cc_env)
+
   def RunWithUnixUtils(cmd, **kwargs):
-    if IsWindows:
+    if IsWindows():
       return Git(['bash'] + cmd, **kwargs)
     else:
       return proc.check_call(cmd, **kwargs)
-  RunWithUnixUtils('ninja', 'check-all'], cwd=LLVM_OUT_DIR, env=cc_env)
+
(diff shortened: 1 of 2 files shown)


dschuff:
Gonna TBR this so I can test it today.

src/build.py
```diff
@@ -783,7 +767,14 @@ def LLVM():
...
   proc.check_call(['ninja', '-v'] + jobs, cwd=LLVM_OUT_DIR, env=cc_env)
-  proc.check_call(['ninja', 'check-all'], cwd=LLVM_OUT_DIR, env=cc_env)
+
+  def RunWithUnixUtils(cmd, **kwargs):
+    if IsWindows():
+      return Git(['bash'] + cmd, **kwargs)
+    else:
+      return proc.check_call(cmd, **kwargs)
+
+  RunWithUnixUtils(['ninja', 'check-all'], cwd=LLVM_OUT_DIR, env=cc_env)
```
jgravelle-google: This looks slightly odd as a local function that only gets used once.
Though it seems like something that can be used elsewhere in the future.
So it probably makes more sense for it to just be a top-level function.

dschuff: Yeah this is really-local and really-hacky (and I hope to keep it from being used anywhere else and hopefully remove it eventually) so I kind of kept it here for that reason.

PR #163, opened 2017-01-12
Run the bare tests on Windows

dschuff:
Several changes:
* Add a no_windows suppression to Test (like for Build)
* Invert the no_windows default for Build since most of the builds work now.
* Support windows in the Musl build
* Support windows in the bare test runners.

### DIFF
--- a/src/assemble_files.py
+++ b/src/assemble_files.py
@@ -34,8 +34,8 @@ def assemble(infile, outfile, extras):
   assembler = extras['assembler']
   basename = os.path.splitext(os.path.basename(assembler))[0]
   commands = {
-      'wast2wasm': [extras['assembler'], infile, '-o', outfile],
-      'wasm-as': [extras['assembler'], infile, '-o', outfile]
+      'wast2wasm': [assembler, infile, '-o', outfile],
+      'wasm-as': [assembler, infile, '-o', outfile]
   }
   return commands[basename]
 
--- a/src/assemble_files.py
+++ b/src/assemble_files.py
@@ -32,7 +32,7 @@ def create_outname(outdir, infile):
 def assemble(infile, outfile, extras):
   """Create the command-line for an assembler invocation."""
   assembler = extras['assembler']
-  basename = os.path.basename(assembler)
+  basename = os.path.splitext(os.path.basename(assembler))[0]
   commands = {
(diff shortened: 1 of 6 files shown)


src/assemble_files.py
```diff
@@ -32,7 +32,7 @@ def create_outname(outdir, infile):
 def assemble(infile, outfile, extras):
   """Create the command-line for an assembler invocation."""
   assembler = extras['assembler']
-  basename = os.path.basename(assembler)
+  basename = os.path.splitext(os.path.basename(assembler))[0]
   commands = {
       'wast2wasm': [extras['assembler'], infile, '-o', outfile],
```
jgravelle-google: Existing nit: should these two `extras['assembler']` references be just `assembler`?

src/build.py
```diff
@@ -1256,10 +1276,10 @@ def TestEmtestAsm2Wasm():
 
 ALL_TESTS = [
     Test('bare', TestBare),
-    Test('asm', TestAsm),
-    Test('emwasm', TestEmwasm),
-    Test('emtest', TestEmtest),
-    Test('emtest-asm', TestEmtestAsm2Wasm),
+    Test('asm', TestAsm, no_windows=True),
```
jgravelle-google: I legitimately like how the default for these is `False`, even though that means more line changes here, because of the consistency with `Source` and `Build`

src/compile_torture_tests.py
```diff
@@ -58,8 +58,6 @@ def run(c, cxx, testsuite, sysroot_dir, fails, out, config='wasm'):
                              'BINARYEN_METHOD="interpret-binary"'],
   }
 
-  assert os.path.isfile(c), 'Cannot find C compiler at %s' % c
```
jgravelle-google: I'm assuming it's ok to not check here because it will fail just fine by itself in the call to compile?

src/assemble_files.py
dschuff: Yeah, done.

src/compile_torture_tests.py
dschuff: Yeah. Although now that I think of it I should put `Executable()` around what we pass in here for consistency with the other tools. I'll do that and then I should be able to put these back.

PR #164, opened 2017-01-12
Use zip archives instead of tar on Windows

dschuff:
The `Tar` function shells out to the `tar` binary, which doesn't exist on Windows. Instead of reimplementing `Tar`, this PR uses a format more "native" to the platform. Also make the naming of the archiving and uploading functions more consistent.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -270,7 +270,7 @@ def UploadFile(local_name, remote_name):
 
 
 def UploadArchive(name, archive):
-  """Archive the tar/zip file with the given name, and with the build number."""
+  """Archive the tar/zip file with the given name and the build number."""
   if not IsBuildbot():
     return
   extension = os.path.splitext(archive)[1]
--- a/src/build.py
+++ b/src/build.py
@@ -218,10 +218,17 @@ def CopyLibraryToArchive(library, prefix=''):
   shutil.copy2(library, install_lib)
 
 
-def Tar(directory, print_content=False):
-  """Create a tar file from directory."""
+def Archive(directory, print_content=False):
+  """Create an archive file from directory."""
+  # Use the format "native" to the platform
   if not IsBuildbot():
     return
+  if IsWindows():
+    return Zip(directory, print_content)
+  return Tar(directory, print_content)
+
+
+def Tar(directory, print_content=False):
   assert os.path.isdir(directory), 'Must tar a directory to avoid tarbombs'
   (up_directory, basename) = os.path.split(directory)
   tar = os.path.join(up_directory, basename + '.tbz2')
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -234,6 +241,26 @@ def Tar(directory, print_content=False):
   return tar
 
 
+def Zip(directory, print_content=False):
+  assert os.path.isdir(directory), 'Must be a directory'
+  (dir, basename) = os.path.split(directory)
```
binji: I think without parens is more canonical

src/build.py
```diff
@@ -218,10 +218,17 @@ def CopyLibraryToArchive(library, prefix=''):
...
-def Tar(directory, print_content=False):
-  """Create a tar file from directory."""
+def Archive(directory, print_content=False):
+  """Create an archive file from directory."""
+  # Use the format "native" to the platform
   if not IsBuildbot():
     return
+  if IsWindows():
+    return Zip(directory, print_content)
+  return Tar(directory, print_content)
```
binji: It's clearer this way, but doesn't it just beg for a sexy one-liner... :)

src/build.py
```diff
@@ -234,6 +241,26 @@ def Tar(directory, print_content=False):
   return tar
 
 
+def Zip(directory, print_content=False):
+  assert os.path.isdir(directory), 'Must be a directory'
+  (dir, basename) = os.path.split(directory)
+  archive = os.path.join(dir, basename + '.zip')
+  print 'Creating zip archive', archive
+  with zipfile.ZipFile(archive, 'w', zipfile.ZIP_DEFLATED) as zip:
+    def Add(path, zippath):
```
binji: Might be a bit clearer with `os.walk`, but this works too

src/build.py
jgravelle-google: `dir` is a python builtin, maybe `dirname`?

src/build.py
```diff
@@ -234,6 +241,26 @@ def Tar(directory, print_content=False):
   return tar
 
 
+def Zip(directory, print_content=False):
+  assert os.path.isdir(directory), 'Must be a directory'
+  (dir, basename) = os.path.split(directory)
+  archive = os.path.join(dir, basename + '.zip')
+  print 'Creating zip archive', archive
+  with zipfile.ZipFile(archive, 'w', zipfile.ZIP_DEFLATED) as zip:
```
jgravelle-google: Ditto with `zip`, maybe `z`?

src/build.py
```diff
@@ -234,6 +241,26 @@ def Tar(directory, print_content=False):
   return tar
 
 
+def Zip(directory, print_content=False):
+  assert os.path.isdir(directory), 'Must be a directory'
+  (dir, basename) = os.path.split(directory)
+  archive = os.path.join(dir, basename + '.zip')
+  print 'Creating zip archive', archive
```
jgravelle-google: Do we want to always print this?
Presumably yes because `Tar` does this via `check_call`

src/build.py
jgravelle-google: return `(Zip if IsWindows() else Tar)(directory, print_content)`!
Simpler cleaner and more inscrutable!

dschuff: Don't toy with me; I will add an unnecessary third argument to either `Zip` or `Tar` to avoid writing `return (Zip if IsWindows() else Tar)(directory, print_content)`

src/build.py
dschuff: Forgot about `os.walk`. Looking at the result, I'm not sure which I prefer but walk is shorter at least?

src/build.py
dschuff: Or better yet, third arguments to each, with different meanings, to tempt the unwary functional programmer and teach them a lesson...

jgravelle-google: Just change the second argument from `print_content` to `dont_print_content` (or `quiet` or w/e), and get around the that pesky type system that would actually alert someone if anything was wrong.

But that's not broken enough, so what about swapping the arguments in one of them, and starting each function with `directory, print_content = str(directory), bool(print_content) #Assert that arguments are correct` to really mess people up.

PR #165, opened 2017-01-12
Only strip off windows exe suffixes in execute_files

dschuff:
`wasm.opt` is the filename of the spec interpreter, so don't strip off
all suffixes; just ones that correspond to Windows executables

### DIFF
--- a/src/execute_files.py
+++ b/src/execute_files.py
@@ -32,7 +32,7 @@ def create_outname(outdir, infile):
 def execute(infile, outfile, extras):
   """Create the command-line for an execution."""
   runner = extras['runner']
-  # Only strip Windows suffxes because wasm.opt has one
+  # Strip only Windows suffxes (instead of all) because wasm.opt has a suffix
   basename = os.path.basename(runner).replace('.exe', '').replace('.bat', '')
   out_opt = ['-o', outfile] if outfile else []
   extra_files = extras['extra_files']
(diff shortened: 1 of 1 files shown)


src/execute_files.py
```diff
@@ -32,7 +32,8 @@ def create_outname(outdir, infile):
 def execute(infile, outfile, extras):
   """Create the command-line for an execution."""
   runner = extras['runner']
-  basename = os.path.splitext(os.path.basename(runner))[0]
+  # Only strip Windows suffxes because wasm.opt has one
+  basename = os.path.basename(runner).replace('.exe', '').replace('.bat', '')
```
binji: could be .cmd too, but I doubt anyone uses those

dschuff: Yeah. After this change (plus my other outstanding ones to finish windows support) I'll probably put some of these into a library, even though the amount of platform-specific stuff ended up being less than I expected.

src/execute_files.py
```diff
@@ -32,7 +32,8 @@ def create_outname(outdir, infile):
 def execute(infile, outfile, extras):
   """Create the command-line for an execution."""
   runner = extras['runner']
-  basename = os.path.splitext(os.path.basename(runner))[0]
+  # Only strip Windows suffxes because wasm.opt has one
```
jgravelle-google: This comment feels incomplete? It reads like "we're only stripping Windows suffixes because wasm.opt has a .exe suffix," which I think is the opposite of what you mean?

dschuff: I guess a better way to say it is "we are stripping only windows suffixes (not all suffixes) because"

PR #166, opened 2017-01-13
Support emscripten on Windows

dschuff:
* Sync the windows node.js
* Build emscripten
* Fix test running and enable asm.js and emwasm tests

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -132,6 +132,10 @@ def Executable(name, extension='.exe'):
   return name + extension if IsWindows() else name
 
 
+def WindowsFSEscape(path):
+  return os.path.normpath(path).replace('\\', '/')
+
+
 # Use prebuilt Node.js because the buildbots don't have node preinstalled
 NODE_VERSION = '7.0.0'
 NODE_BASE_NAME = 'node-v' + NODE_VERSION + '-'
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -545,7 +569,7 @@ def NoSync(*args):
     Source('cmake', '', '',  # The source and git args are ignored.
            custom_sync=SyncPrebuiltCMake),
     Source('nodejs', '', '',  # The source and git args are ignored.
-           custom_sync=SyncPrebuiltNodeJS, no_windows=True),
+           custom_sync=SyncPrebuiltNodeJS, no_windows=False),
```
jgravelle-google: Should just remove `no_windows=False`

src/build.py
```diff
@@ -140,13 +142,14 @@ def Executable(name):
 
 
 def NodePlatformName():
-  if IsWindows():
-    return ''
-  return {'darwin': 'darwin-x64', 'linux2': 'linux-x64'}[sys.platform]
+  return {'darwin': 'darwin-x64',
+          'linux2': 'linux-x64',
+          'win32':'win32'}[sys.platform]
```
jgravelle-google: nit: space after `:` (flake8 doesn't care?)

src/build.py
```diff
@@ -1170,7 +1195,7 @@ def AllBuilds(use_asm=False):
       Build('spec', Spec, no_windows=True),
       Build('binaryen', Binaryen),
       Build('fastcomp', Fastcomp),
-      Build('emscripten', Emscripten, True, use_asm),
+      Build('emscripten', Emscripten, False, use_asm),
```
jgravelle-google: We should pass these as kwargs

src/build.py
dschuff: Done. Eventually I hope to remove `no_windows` entirely. Although it's likely we won't want to run every test or step on every platform so we may want something more general instead.

src/build.py

src/build.py
dschuff: (already done, because flake8 does care)

PR #167, opened 2017-01-13
Fixes for fastcomp and binaryen windows builds

dschuff:
Binaryen debug build is pathologically slow on Windows, so use Release
Install MSVC dlls at fastcomp install time (after the dest dir exists), not separately.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -886,9 +886,10 @@ def Binaryen():
 
   proc.check_call(
       [PREBUILT_CMAKE_BIN, '-G', 'Ninja', BINARYEN_SRC_DIR,
+       '-DCMAKE_BUILD_TYPE=Release',
        '-DCMAKE_INSTALL_PREFIX=%s' % INSTALL_DIR] + OverrideCMakeCompiler(),
       cwd=BINARYEN_OUT_DIR, env=cc_env)
-  proc.check_call(['ninja'], cwd=BINARYEN_OUT_DIR, env=cc_env)
+  proc.check_call(['ninja', '-v'], cwd=BINARYEN_OUT_DIR, env=cc_env)
   proc.check_call(['ninja', 'install'], cwd=BINARYEN_OUT_DIR, env=cc_env)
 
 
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -886,9 +886,10 @@ def Binaryen():
 
   proc.check_call(
       [PREBUILT_CMAKE_BIN, '-G', 'Ninja', BINARYEN_SRC_DIR,
+       '-DCMAKE_BUILD_TYPE=Release',
        '-DCMAKE_INSTALL_PREFIX=%s' % INSTALL_DIR] + OverrideCMakeCompiler(),
       cwd=BINARYEN_OUT_DIR, env=cc_env)
-  proc.check_call(['ninja'], cwd=BINARYEN_OUT_DIR, env=cc_env)
+  proc.check_call(['ninja', '-v'], cwd=BINARYEN_OUT_DIR, env=cc_env)
```
jgravelle-google: I assume the `-v` was added for debugging while the change was being made, but I also assume that it'd be a useful thing to have in perpetuity.

src/build.py
```diff
@@ -886,9 +886,10 @@ def Binaryen():
 
   proc.check_call(
       [PREBUILT_CMAKE_BIN, '-G', 'Ninja', BINARYEN_SRC_DIR,
+       '-DCMAKE_BUILD_TYPE=Release',
```
jgravelle-google: This changes the build from debug to release on all platforms?
Seems fine.

src/build.py
```diff
@@ -911,6 +912,8 @@ def Fastcomp():
        '-DLLVM_ENABLE_ASSERTIONS=ON'] + OverrideCMakeCompiler(),
       cwd=FASTCOMP_OUT_DIR, env=cc_env)
   proc.check_call(['ninja'], cwd=FASTCOMP_OUT_DIR, env=cc_env)
+  # Fastcomp has a different install location than the rest of the tools
+  BuildEnv(install_dir, bin_subdir=True)
```
jgravelle-google: Seems mildly smelly to call this just for its side effects, but I think the pragmatic factoring is pretty much just this, so nbd

src/build.py
dschuff: Yeah, currently Binaryen's `CMakeLists.txt` currently enables asserts for Release builds on Linux but not Windows; something I'm about to submit a PR to fix. With the asserts I'm fine making it Release everywhere.

src/build.py
dschuff: Yeah, in general IMO bots should be as verbose as possible as long as the amount of output is manageable.

src/build.py
dschuff: Yeah. Also if we kept the convention of "function names are always verbs" this would be more clear too, as this function both builds the runtime environment (i.e. the dlls) for the binaries, and returns environment vars to inject into the build environment.

PR #172, opened 2017-02-15
Add jsc

jfbastien:


src/test/jsc_musl_known_gcc_test_failures.txt
```diff
@@ -0,0 +1,93 @@
+# Expected failures from running the assembled GCC torture tests with d8.
```
binji: jsc

src/test/jsc_known_gcc_test_failures.txt
```diff
@@ -0,0 +1,108 @@
+# Expected failures from running the assembled GCC torture tests with d8.
```
binji: jsc

jfbastien:
@binji good catch. Done.

dschuff:
Cool, I'm going to try this locally.

src/build.py
```diff
@@ -840,6 +858,27 @@ def V8():
     CopyBinaryToArchive(os.path.join(V8_OUT_DIR, a))
 
 
+def Jsc():
+  buildbot.Step('JSC')
+  Mkdir(JSC_OUT_DIR)
+
+  proc.check_call(['xcrun', PREBUILT_CMAKE_BIN, '-Wno-dev',
```
dschuff: This might potentially be a problem because of the way that dev tools are "installed" on Chromium infrastructure; i.e. they are not really installed per se on the system, so sometimes assumptions about the installation are wrong, etc. Practically it may just be that we need an explicit path to xcrun or to add something to the $PATH or something. I'm trying it out now, will let you know

src/build.py
```diff
@@ -840,6 +858,27 @@ def V8():
...
+                  cwd=JSC_OUT_DIR)
+  proc.check_call(['ninja', 'jsc'], cwd=JSC_OUT_DIR)
+  proc.check_call(['../Tools/Scripts/run-javascriptcore-tests',
+                   '--root=bin',
+                   '--filter', 'wasm',
+                   '--no-build', '--no-testapi'],
+                  cwd=JSC_OUT_DIR)
+  to_archive = [Executable(os.path.join('bin', 'jsc'))]
+  for a in to_archive:
+    CopyBinaryToArchive(os.path.join(JSC_OUT_DIR, a))
```
dschuff: any idea approximately how big this output is? Is it just the jsc binary, or does it need supporting files?

src/build.py
```diff
@@ -840,6 +858,27 @@ def V8():
...
+  Mkdir(JSC_OUT_DIR)
+
+  proc.check_call(['xcrun', PREBUILT_CMAKE_BIN, '-Wno-dev',
+                   '..', '-G', 'Ninja',
+                   '-DCMAKE_BUILD_TYPE="Release"',
+                   '-DPORT=Mac',
+                   '-DENABLE_WEBASSEMBLY=ON'],
+                  cwd=JSC_OUT_DIR)
+  proc.check_call(['ninja', 'jsc'], cwd=JSC_OUT_DIR)
+  proc.check_call(['../Tools/Scripts/run-javascriptcore-tests',
```
dschuff: Oh and it would be great if we could use goma here too. Care to refactor the goma+cmake logic out of the LLVM build rule?

src/build.py
jfbastien: Just the jsc binary, which is 315K in release mode.

src/build.py
jfbastien: Sure, I'm happy to update with whatever.

src/build.py
jfbastien: It runs in about 20s. I'm not sure goma will save time?

dschuff: The build was 1117 files on my laptop. Surely goma would be faster than compiling that with -j4?

jfbastien: Oh the build? You put the comment on the test line 😛 
I'll take a look soon.

dschuff: yeah sorry I fail at clicking.

jfbastien: So this part:
```

  if 'GOMA_DIR' in os.environ:
    compiler_launcher = os.path.join(os.environ['GOMA_DIR'], 'gomacc')
    jobs = ['-j', '50']
  else:
    try:
      compiler_launcher = proc.Which('ccache', WORK_DIR)
      command.extend(['-DCMAKE_%s_FLAGS=-Qunused-arguments' %
                      c for c in ['C', 'CXX']])
    except:
      compiler_launcher = None
```
?

Can you test that it works with the cmake + xcrun magic thing?

dschuff: Yeah. I expect that it will because it's a separate thing that CMake just prepends. But I also need to comandeer the bot slave itself and try it (or just merge and see if it works) because unlike the bot slave, I actually do have a real install of XCode on my machine.

src/build.py
jfbastien: Sooooo... How did your experiment work out?

src/build.py
jfbastien: Done. It could do with some factoring out, which I'll leave as an exercise to the reader.

jfbastien:
@dschuff ping.

dschuff:
OK I tested this again locally and it now has exactly the same unexpected failures as d8 (and that's something we are looking into shortly) so whatever triage or fix we do for that we can just apply here as well. So I think this can just go in, and we can iterate.

jfbastien:
It is done! 🎉

dschuff:
...aaaand, it's broken 💣
https://wasm-stat.us/builders/mac/builds/3570/steps/Sync%20Repos/logs/stdio

```
subprocess.check_call(`/usr/local/git/bin/git clone https://github.com/WebKit/webkit /b/build/slave/mac/build/src/src/work/jsc --depth 1000`, cwd=`/b/build/slave/mac/build/src/src`)
Cloning into '/b/build/slave/mac/build/src/src/work/jsc'...
remote: Not Found
fatal: repository 'https://github.com/WebKit/webkit/' not found
```
weird...

dschuff:
wait, should that be `https://github.com/WebKit/webkit.git`? And if so, how did it ever work for me locally?

jfbastien:
🤷 ¯\_(ツ)_/¯ 🤷

sbc100:
I don't think you need the .git extension when checking out github repos.. at least not from the command line.

PR #173, opened 2017-03-13
Move prebuilt clang checkout to root of working dir

dschuff:
A recent change
(https://chromium.googlesource.com/chromium/src/tools/clang/+/49cece5b02431a848652c7f60dd7b06d94e8d625)
made the prebuilt gold plugin downloaded by default, and also made the
location of the checkout dependent on the location of the binaries. So
move the checkout out to the root of the working dir to make things line
up.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -63,7 +63,7 @@ MUSL_SRC_DIR = os.path.join(WORK_DIR, 'musl')
 
 FIND_SVN_REV = os.path.join(SCRIPT_DIR, 'find_svn_rev.py')
 
-PREBUILT_CLANG = os.path.join(WORK_DIR)
+PREBUILT_CLANG = WORK_DIR
 PREBUILT_CLANG_TOOLS_CLANG = os.path.join(PREBUILT_CLANG, 'tools', 'clang')
 PREBUILT_CLANG_BIN = os.path.join(
     PREBUILT_CLANG, 'third_party', 'llvm-build', 'Release+Asserts', 'bin')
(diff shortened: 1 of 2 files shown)


dschuff:
@jgravelle-google should fix the linux build.

src/build.py
```diff
@@ -63,7 +63,7 @@
 
 FIND_SVN_REV = os.path.join(SCRIPT_DIR, 'find_svn_rev.py')
 
-PREBUILT_CLANG = os.path.join(WORK_DIR, 'chromium-clang')
+PREBUILT_CLANG = os.path.join(WORK_DIR)
```
sbc100: This join function call is now redundant, no?

dschuff: yes, removed.

dschuff:
So the upstream change making the gold plugin unconditionally downloaded was reverted, so we probably don't need this. But the wabt build on Travis failed in a way that looks unrelated. @binji this looks like the printf stuff we've been messing with lately, any idea what the issue is here?

binji:
@dschuff My guess is that we'll have to unconditionally define `__STDC_FORMAT_MACROS`. It was necessary for mingw, but not for any other compiler. Perhaps because that was the newest version of gcc? Similarly here, it looks like Clang is quite new.

dschuff:
Ah yeah, this build uses chromium's clang, which is quite new.

dschuff:
Although actually if that's the only problem, then it seems like the wabt build would also be failing on https://wasm-stat.us and it isn't...

binji:
True, though I'm having trouble finding a build where it is actually building anything, since it's an incremental build. I guess if it were failing it would continue to fail, though.

dschuff:
Yeah, maybe it's the libc headers on Travis vs on chromium infrastructure.

dschuff:
This isn't needed anymore, closing.

PR #174, opened 2017-03-15
Fix URL of webkit git repo

dschuff:


jfbastien:
I'm baffled it ever worked.

PR #175, opened 2017-03-16
Make syncing and testing JSC also Mac-only

jgravelle-google:


jfbastien:
This should also work on Linux and FreeBSD FWIW. You don't need to restrict it to Mac.

dschuff:
Well the build of JSC is restricted to mac already, no?

jfbastien:
@dschuff correct, I'm saying it doesn't need to be anymore. No Windows, but Linux works.

PR #177, opened 2017-03-24
Protects build from failing when JSC is not compiled.

yurydelendik:
The "mac" builder does not reach "Archive binaries" step after #172.
Allows execution of more steps even if JSC cannot be built.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -899,6 +899,14 @@ def Jsc():
     CopyBinaryToArchive(os.path.join(JSC_OUT_DIR, a))
 
 
+def JscSafe():
+  try:
+    Jsc()
+  except proc.CalledProcessError:
+    # Note the failure but allow the build to continue.
+    buildbot.Fail()
+
+
 def Wabt():
   buildbot.Step('WABT')
   Mkdir(WABT_OUT_DIR)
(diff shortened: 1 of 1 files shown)


jfbastien:
Does this generate LKGR, or does it simply archive without marking the build as successful?

jfbastien:
As long as it doesn't get tagged as LKGR, I don't really care what it runs or what gets archived.

jfbastien:
FWIW I just fixed the test failure.

dschuff:
FWIW we've updated the mac bot to a newer version (this is what was causing the build failure previously; it was working fine on my laptop). Now we're getting a new build failure, one that I can reproduce on my laptop (https://wasm-stat.us/builders/mac/builds/3916/steps/JSC/logs/stdio). It seems to have been introduced in https://github.com/WebKit/webkit/commit/01a033c6278902a7cc297bec2ab66db714a5057d
JF, does this ring any bells or do you have any idea about it? Otherwise, where should I report a bug? I assume this arises from some difference in an "internal" vs "external" style build, or some difference in the SDK used (I tried with 12.12 and 12.11 SDKs) else maybe it would have been caught already?

jfbastien:
@dschuff cmake isn't the official build, but it's way faster than the make build. It sometimes breaks, a few of us notice and fix it. This seems like it's due to a file deletion / rename. For these I usually just grep for the file name, it finds the recent ChangeLog entry, and then the fix is obvious.

I'll take a look in the next few hours, once baby is away, and if it's not yet fixed I'll do so.

jfbastien:
@dschuff seems fixed now.

dschuff:
@jfbastien thanks, JSC is building now. Next step: make waterfall green again!

PR #178, opened 2017-03-27
Toggle alloca-1.c torture test for asm2wasm, add comment explaining it

jgravelle-google:


jgravelle-google:
Test context: https://github.com/gcc-mirror/gcc/blob/master/gcc/testsuite/gcc.c-torture/execute/alloca-1.c

It looks like a gcc builtin that isn't fully supported in Clang (https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html).

PR #181, opened 2017-04-05
Implement test attributes for known failure files

dschuff:
Augment the syntax used in test exclusion ('known failure') files to use attributes for torture test run configurations and exclusions. This allows most of the test exclusions to be combined into one file, where they can use common explanations, and where it's easier to see which combinations of conditions cause the failures.
This PR combines the bare JS engine and emscripten test exclusions into a single file, but not the spec interpreter (it has a large set and most failures are just due to missing imports; we may try it in the future) or the Binaryen shell (it lives in a different repo), or the compile failures (which are of course disjoint from the runtime failures).


dschuff:
@jfbastien I noticed while combining these that jsc-musl seems broken with an internal assertion failure: https://wasm-stat.us/builders/mac/builds/4040/steps/Execute%20LLVM%20Torture%20%28jsc-musl%29/logs/stdio

jfbastien:
@dschuff should be fixed: https://bugs.webkit.org/show_bug.cgi?id=170462

src/testing.py
```diff
@@ -89,6 +89,46 @@ def __call__(self, test_file):
...
+  the test configurations which are expected to fail. (e.g. JS engine
+  or optimization level). A test is only excluded if the configuration
+  has all the attributes specified in the exclude line. Lines which
+  have no attributes will match everything, and lines which specify only
+  one attribute (e.g. engine) will match all configurations with that
+  attribute (e.g. both opt levels with that engine). Returns a sorted list
+  of exclusions which match the attributes.
+  '''
+  excludes = {}  # maps name of excluded test to file from whence it came
+  if not config_attributes:
```
jgravelle-google: Would a default argument make sense instead?

Microoptimization: I feel like this should be checking `is None`, because if you call this with an empty set or array, we'll allocate a new empty set, which is work we don't need to do.

src/testing.py
```diff
@@ -89,6 +89,46 @@ def __call__(self, test_file):
       return Result(test=basename, success=False, output=e.output)
 
 
+def parse_exclude_files(fails, config_attributes):
+  ''' Parse the files containing tests to exclude (i.e. expected fails).
```
jgravelle-google: This is kind of a block of text. How about a bulleted list, with nesting for the examples? I realize there is only so much to be done within 79 characters, but the interleaved parenthetical statements made it hard to read at first.

[PEP257](https://www.python.org/dev/peps/pep-0257/) nits:
* First line should be an executive summary, followed by a blank line
* Docstrings should use triple-doubles

Quick reformatting:
```
"""Returns a sorted list of exclusions which match the attributes.

Parse the files containing tests to exclude (i.e. expected fails).
* Each line may contain a comma-separated list of attributes restricting
  the test configurations which are expected to fail.
    - e.g. JS engine or optimization level.
* A test is only excluded if the configuration has all the attributes
  specified in the exclude line.
* Lines which have no attributes will match everything, and lines which
  specify only one attribute (e.g. engine) will match all configurations
  with that attribute (e.g. both opt levels with that engine).
"""
```

Given that, the last bullet feels like it can be reworded to split into two, which lets the parentheticals drop down more naturally.

Also given that we're many lines in at this point, some example lines might be useful

src/testing.py
```diff
@@ -89,6 +89,46 @@ def __call__(self, test_file):
...
+    config_attributes = set()
+  for excludefile in fails:
+    f = open(excludefile)
+    for line in f:
+      line = line.strip()
+      if '#' in line:
+        line = line[:line.index('#')].strip()
+      if not line:
+        continue
+      tokens = line.split()
```
jgravelle-google: Getting tokens from a line feels like it could be its own function. Change `if not line` to `if not tokens` for the `continue` and it should work.

src/testing.py
```diff
@@ -89,6 +89,46 @@ def __call__(self, test_file):
...
+      if len(tokens) > 1:
+        attributes = set(tokens[1].split(','))
+        if not attributes.issubset(config_attributes):
+          continue
+        test = tokens[0]
+      else:
```
jgravelle-google: The control flow here feels hairy. I feel like we want something more like:
```python
test, attributes = get_test_and_attributes_from(line)
if not attributes.issubset(config_attributes):
  continue
```

Alternatively, moving `test = tokens[0]` above `attributes = ...` makes it clearer that we're `continue`ing to avoid adding `test` to `excludes`; as it is now it looks more like we want to avoid reading the first token.

src/testing.py
```diff
@@ -89,6 +89,46 @@ def __call__(self, test_file):
...
+        test = line
+      if test in excludes:
+        print 'ERROR: duplicate exclude: [%s]' % line
+        print 'Files: %s and %s' % (excludes[test], excludefile)
+        sys.exit(1)
```
jgravelle-google: Is this a hard error?
What if we wanted something like
```
# foo.c only works with -O0
foo.c o1
foo.c o2
```

Should we add a magic `known-dupe` attribute? (no we should not)
Should it be a warning? (we'll still want the attribute to silence the warnings I'll bet)
Should we not care? (we'll have duplicates and not know)
I vote we don't care, because when the status changes and we go to remove one we'll see it still isn't running.
Turned around, what's the reason we want to disallow duplicate excludes?

src/testing.py
```diff
@@ -89,6 +89,46 @@ def __call__(self, test_file):
...
+      excludes[test] = excludefile
+    f.close()
+    print 'Size of excludes now: %d' % len(excludes)
```
jgravelle-google: Debug print?

src/testing.py
```diff
@@ -89,6 +89,46 @@ def __call__(self, test_file):
...
+  has all the attributes specified in the exclude line. Lines which
+  have no attributes will match everything, and lines which specify only
+  one attribute (e.g. engine) will match all configurations with that
+  attribute (e.g. both opt levels with that engine). Returns a sorted list
+  of exclusions which match the attributes.
+  '''
+  excludes = {}  # maps name of excluded test to file from whence it came
+  if not config_attributes:
+    config_attributes = set()
+  for excludefile in fails:
```
jgravelle-google: Why is `fails` a list of known failure files? I don't see a path where we call this with more than one file currently.
Why would we want to split the known failures into multiple files in the future, when we can slice tests with attributes?

src/testing.py
dschuff: Done.

src/testing.py
dschuff: It's been simplified with pulling the test parsing out, PTAL.

src/build.py
```diff
@@ -1277,11 +1268,13 @@ def TestBare():
       assembler=Executable(os.path.join(INSTALL_BIN, 'wast2wasm')),
       indir=s2wasm_out,
       fails=WAST2WASM_KNOWN_TORTURE_FAILURES)
+  common_attrs = ['bare', 'O0']
```
jgravelle-google: That sounds both pythonic and resilient, nice. In my head I was thinking "let's just remember to always pass sets in", which now that I say out loud is a vastly worse idea. :)

src/testing.py
jgravelle-google: Oh I see. I was confused because we only add the test file without the attributes to `excludes`, and I had paged out the `continue` that means we only do so after filtering based on attributes.
So, seems like something we should enforce 100% of the time and I'm glad it's here!

src/testing.py
jgravelle-google: That's much nicer, thanks!

src/testing.py
```diff
@@ -89,6 +89,46 @@ def __call__(self, test_file):
...
+    for line in f:
+      line = line.strip()
+      if '#' in line:
+        line = line[:line.index('#')].strip()
+      if not line:
+        continue
+      tokens = line.split()
+      if len(tokens) > 1:
+        attributes = set(tokens[1].split(','))
+        if not attributes.issubset(config_attributes):
```
jgravelle-google: I should have been more explicit that "I just now learned about set operators, and now let's all quietly forget about them" :)

But yeah, I like the symmetry of both objects being sets, if ever we want to do set-operations with `config_attributes`. It's really clear now that it's always a set, which cuts down lots of future code reading.

src/testing.py
jgravelle-google: Yeah it seemed like a feature PNaCl was using, but I wasn't able to guess what.
Sounds good though. I think there's a bunch of "this fails in native so it's just a clang thing don't worry about it" tests that that sounds perfect for.

src/testing.py
jgravelle-google: Agree. I think my real complaint before was you'd have something like:
```
input  |  output
-------+------------
None   |  new {}
[]     |  new {}
[1, 2] |  old [1, 2]
{}     |  new {}
{1, 2} |  old {1, 2}
```
whereas now it looks more like
```
input  |  output
-------+------------
None   |  new {}
[]     |  new {}
[1, 2] |  new {1, 2}
{}     |  new {}
{1, 2} |  new {1, 2}
```
which is way more predictable.

src/testing.py
jgravelle-google: Any thoughts on the docstring?

dschuff: Reformatted as suggested, with reference to the shared exclusion file which has a longer explanation.

PR #184, opened 2017-04-06
Update Travis build to use Trusty for a newer C++ compiler

dschuff:
The WABT build uses C++11 now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-sudo: false
+sudo: required
+dist: trusty
 language: python
 addons:
   apt:


.travis.yml
```diff
@@ -1,4 +1,5 @@
-sudo: false
+sudo: required
+dist: trusty
```
sbc100: According to this you might not need sudo anymore:
https://docs.travis-ci.com/user/trusty-ci-environment/

dschuff: Oh sweet, they updated. Just in time to be a year late for Xenial :)

PR #186, opened 2017-04-12
Test llvm backend with -c in addtion to -S

sbc100:
Previously we were only testing with -S and using s2wasm
for linking.  Going forward we want to test with -c and
eventuaully with lld and o2wasm.


jfbastien:
Toture tests are single-file. Does this also work with multi-file? Maybe the LLVM test suite would be a good supplement.

src/compile_torture_tests.py
```diff
@@ -82,7 +89,7 @@ def run(c, cxx, testsuite, sysroot_dir, fails, out, config='wasm'):
   return result
 
 
-def getargs():
```
jgravelle-google: Why not keep `getargs`? We can still wrap the call to `run` in a `main` function, but I think having one function do one thing is a useful property to keep.

src/build.py
```diff
@@ -161,8 +162,11 @@ def NodePlatformName():
 
 # Known failures.
 IT_IS_KNOWN = 'known_gcc_test_failures.txt'
-LLVM_KNOWN_TORTURE_FAILURES = [os.path.join(LLVM_SRC_DIR, 'lib', 'Target',
-                                            'WebAssembly', IT_IS_KNOWN)]
+LLVM_KNOWN_TORTURE_FAILURES_S = [os.path.join(LLVM_SRC_DIR, 'lib', 'Target',
+                                              'WebAssembly', IT_IS_KNOWN)]
+# TODO(sbc): Move this failures file to the llvm repo
```
jgravelle-google: *Should* this get moved to llvm? I'd assume upstream llvm wouldn't care about the gcc torture tests, so this seems like it could just live in the waterfall repo.

dschuff: Its counterpart (the known compile failures with -s) is in the LLVM repo, in the wasm backend directory; they should at least be together. The idea behind putting them in the LLVM repo was that you could fix a compile failure and update the expectations in a single commit. [edit: but see below, we probably don't need a separate file]

src/test/wasm_o_compile_known_gcc_test_failures.txt
```diff
@@ -0,0 +1,86 @@
+# Tests which are known to fail from the GCC torture test suite.
```
dschuff: I think this is a good candidate for using the capability introduced in #181. We could run builds with e.g. `assembly` and `object` attributes; most of the fails here would be unconditional (they fail for both), and then the few that are only here would have the `object` attribute on the failure entry.

src/build.py
sbc100: As long as they are on the same place I don't really care which it is.

src/compile_torture_tests.py
sbc100: I guess I'm just used to seeing 'main' do the argument handling since nobody else needs to do this.
 
If I do it as you suggest main becomes two line function and the only caller to getargs()... so I guess like it better this way but its not a huge deal.

sbc100:
@jfbastien, no, we don't currently have a linker story.  But the plan is to add that next.

1. Add this phase that tests compiling to .o (.wasm) files
2.a. Add lld linking phase that handles linking single file .o files into runnable .wasm files.
2.b. Add o2wasm linking (like we already do with s2wasm but with these .o files)
3. Add multi-file linking, and possibly some kind of libc support.

2.a. might still be a ways off as none of it has landed in lld yet (https://github.com/sbc100/lld/pull/2)

jfbastien:
OK. Do you know if the LLVM test suite single file tests work? If so, can a follow-up patch add them to the waterfall?

sbc100:
I haven't yet tried the LLVM test suite single file tests (my lld prototype is still in the early stages), but will certainly add those as soon as they work.

jfbastien:
> I haven't yet tried the LLVM test suite single file tests (my lld prototype is still in the early stages), but will certainly add those as soon as they work.

Do you need lld to make them work? The C library functionality can all be imported from JS.

sbc100:
I'll give it a go.  Its not clear that making the .o files directly runnable was a goal but perhaps it would be do-able.

src/compile_torture_tests.py
sbc100: Reverted this an created a separate PR for bikeshedding

jgravelle-google: I think if `main` were doing more things I'd feel more strongly about it. As it is `main` gets the arguments, delegates to the runner, and returns the exit code.

And +1 to consistency everywhere in https://github.com/WebAssembly/waterfall/pull/188

src/test/wasm_o_compile_known_gcc_test_failures.txt
sbc100: Done! Thanks.

src/test/wasm_compile_known_gcc_test_failures.txt
```diff
@@ -0,0 +1,98 @@
+# Tests which are known to fail from the GCC torture test suite.
+# TODO(sbc): The original home for this file is in the llvm repo once that
```
dschuff: I believe nothing is stopping you from just updating the LLVM repo first (to exactly the contents you have here) before landing this PR.

sbc100: Done (once you review https://reviews.llvm.org/D32300)

sbc100:
Should be good now.  PTAL

PR #187, opened 2017-04-13
Switch llvm mirror to https://llvm.googlesource.com/

sbc100:


src/build.py
```diff
@@ -100,9 +100,8 @@
 GITHUB_REMOTE = 'github'
 GITHUB_SSH = 'git@github.com:'
 GIT_MIRROR_BASE = 'https://chromium.googlesource.com/'
-LLVM_OFFICIAL_MIRROR_BASE = GIT_MIRROR_BASE + 'external/llvm.org/'
+LLVM_MIRROR_BASE = 'https://llvm.googlesource.com/'
 GITHUB_MIRROR_BASE = GIT_MIRROR_BASE + 'external/github.com/'
```
dschuff: I think we can also remove this line now too.

sbc100: Its used in the lines below isn't it?

dschuff: oh, right. yeah i was just looking below.

jfbastien:
I love the proliferation of git mirrors.

sbc100:
I'm trying to reduce it .. hopefully we can rid of these ones that I'm deleting here.

PR #192, opened 2017-04-26
Add --no-wasm-async-compilation when running d8 tests

sbc100:
Async compilation is currently breaking the d8 tests
because its exiting before the promises resolve.

BUG=https://bugs.chromium.org/p/v8/issues/detail?id=6263


dschuff:
oh, you should probably also add a comment with a link to https://bugs.chromium.org/p/v8/issues/detail?id=6263

PR #200, opened 2017-05-18
Test prototype lld linker

sbc100:
Temporarily build lld from sbc's branch (since it is not
yet upstream).  Test linking and executing the resulting
binary for bare tests.


src/test/run_known_gcc_test_failures.txt
```diff
@@ -20,8 +20,12 @@
 # come with 'bare' or 'bare-musl' attributes (based on whether they use the
 # hacky JS libc or the hacky linking of musl). The failures are mostly the same
 # for those. These tests also are run with d8 and JSC ('d8' and 'jsc' attributes).
+#
 # Files with a .js extension are emscripten builds, which come in 'asm2wasm' or
 # 'emwasm' (aka LLVM upstream wasm backend) attributes.
+#
+# Files with a .o.wasm extension are build using clang and lld native backends.
```
jgravelle-google: Typo: are buil**t**

src/link_assembly_files.py
```diff
@@ -23,20 +23,27 @@
...
 
 def link(infile, outfile, extras):
   """Create the command-line for a linker invocation."""
   linker = extras['linker']
   basename = os.path.splitext(os.path.basename(linker))[0]
+  install_root = os.path.dirname(os.path.dirname(linker))
+  symfile = os.path.join(install_root, 'lib', 'syms.def')
   commands = {
+      'lld': [linker, '-flavor', 'wasm', '-entry=main',
+        '--allow-undefined-file='+ symfile, '-o', outfile, infile],
```
jgravelle-google: Nit: Should probably align the whitespace of the second line to the `[`, like

```python
'lld': [linker, '-flavor', 'wasm', '-entry=main',
        '--allow-undefined-file='+ symfile, '-o', outfile, infile],
```

Nit: the spacing around the `+` isn't consistent.

src/link_assembly_files.py
```diff
@@ -23,20 +23,27 @@
...
+    outname = basename + '.wast'
   return os.path.join(outdir, outname)
 
 
 def link(infile, outfile, extras):
   """Create the command-line for a linker invocation."""
   linker = extras['linker']
   basename = os.path.splitext(os.path.basename(linker))[0]
+  install_root = os.path.dirname(os.path.dirname(linker))
+  symfile = os.path.join(install_root, 'lib', 'syms.def')
```
jgravelle-google: Is there a more descriptive name we can use than "syms.def"? I'm not sure whether these are test-only symbols or something else.

src/compile_torture_tests.py
```diff
@@ -28,18 +28,10 @@ def c_compile(infile, outfile, extras):
...
-  So we manually package the suffix in a class.
-  """
-  def __init__(self, suffix):
-    self.suffix = suffix
-
-  def __call__(self, outdir, infile):
-    basename = os.path.basename(infile)
-    outname = basename + self.suffix
-    return os.path.join(outdir, outname)
+def create_outname(outdir, infile, extras):
```
jgravelle-google: Nice

src/build.py
```diff
@@ -1272,18 +1279,41 @@ def Test(self):
...
   CompileLLVMTorture('s', TORTURE_S_OUT_DIR)
   CompileLLVMTorture('o', TORTURE_O_OUT_DIR)
+
+  # Link/Assemble
+  lld_out = LinkLLVMTorture(
+      name='lld',
+      linker=Executable(os.path.join(INSTALL_BIN, 'lld')),
+      fails=LLD_KNOWN_TORTURE_FAILURES,
+      input_pattern=os.path.join(TORTURE_O_OUT_DIR, '*.o'))
+
```
jgravelle-google: Nit: Inconsistent spacing for lld, s2wasm, and wasm2wast. I have no strong opinion whether there should be spaces between or not, but they should match each other within a section.

sbc100:
Fixed everything.  Also add a musl build based on .o files now that libc.py in musl supports it.

PR #202, opened 2017-06-08
Use up-to-date build\{vs_toolchain.py,win_toolchain.json}

rongjiecomputer:
`v8\gypfiles` is deprecated and will be removed completely end of 2017. Use updated scripts in `build` folder instead.

### DIFF
--- a/src/host_toolchains.py
+++ b/src/host_toolchains.py
@@ -31,6 +31,7 @@ WIN_TOOLCHAIN_JSON = os.path.join(CR_BUILD_DIR, 'win_toolchain.json')
 
 os.environ['GYP_MSVS_VERSION'] = '2015'
 
+
 def SyncPrebuiltClang(name, src_dir, git_repo):
   """Update the prebuilt clang toolchain used by chromium bots"""
   tools_clang = os.path.join(src_dir, 'tools', 'clang')
--- a/src/host_toolchains.py
+++ b/src/host_toolchains.py
@@ -26,8 +26,8 @@ CR_BUILD_DIR = os.path.join(WORK_DIR, 'build')
 SETUP_TOOLCHAIN = os.path.join(CR_BUILD_DIR, 'toolchain', 'win',
                                'setup_toolchain.py')
 V8_SRC_DIR = os.path.join(WORK_DIR, 'v8', 'v8')
-VS_TOOLCHAIN = os.path.join(V8_SRC_DIR, 'gypfiles', 'vs_toolchain.py')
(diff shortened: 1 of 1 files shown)


rongjiecomputer:
Travis failed, but does not seem to be related to my changes.

Unrelated comments:

- https://wasm-stat.us/console still points to https://chromium.googlesource.com/external/github.com/llvm-mirror/llvm
- All bots on https://wasm-stat.us/console are red/orange.

sbc100:
@binji  looks like its the old version re2c on travis.   Can we detect the old version and fall back to prebuilt?

binji:
Hm, maybe. Let me see.

binji:
See https://github.com/WebAssembly/wabt/pull/485.

rongjiecomputer:
@binji Travis check passed.

src/host_toolchains.py
```diff
@@ -26,8 +26,8 @@
 SETUP_TOOLCHAIN = os.path.join(CR_BUILD_DIR, 'toolchain', 'win',
                                'setup_toolchain.py')
 V8_SRC_DIR = os.path.join(WORK_DIR, 'v8', 'v8')
```
sbc100: Looks like you can remove this line too.

Do you know why v8 would have been used to get these files in the first place?

rongjiecomputer: Done.

When V8 migrated from GYP to GN last year, `v8\gypfiles` and some `.gyp` files was not removed because some external embedders like Node.js still depend on GYP. However, V8 team decided that maintaining both GYP and GN build together is too much burden, so end of this year all GYP files will be removed from V8. The burden to maintain GYP is shifted to Node.js.

sbc100: I guess my question could have been more clear: Why did the original author every think it sue vs_toolchain.py from v8 rather than chromium (which would seem to be the more logically choice since chromium is effectively the upstream for these files).   Did the v8 version have some useful difference?  I think probably not but just curious really.

rongjiecomputer: > Did the v8 version have some useful difference?

As far as I know, there is no useful difference, just out-dated.

I believe this out-dated vs_toolchain.py is pulling the older version of VS 2015, so when `ninja` spawns 50 jobs to build LLVM, 50 of them fail with compile error (see [here](https://wasm-stat.us/builders/windows/builds/8716/steps/LLVM/logs/stdio/text))

dschuff: The original author was trying to get away without having the cr-build-tools repo at all, and just using v8. But now that we depend on that anyway, I don't have any objection to this PR.

rongjiecomputer:
No more objection?

PR #205, opened 2017-06-12
Make running of llvm/v8/spec tests optional

sbc100:
These tests are normally run as part of the build but
it can be useful to run a build with no tests too.

Alternatively we could split out these tests into
thier own phases.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -133,6 +133,7 @@ GNUWIN32_ZIP = 'gnuwin32.zip'
 
 options = None
 
+
 def IsWindows():
   return sys.platform == 'win32'
 
--- a/src/build.py
+++ b/src/build.py
@@ -131,6 +131,7 @@ OCAML_BIN_DIR = os.path.join(OCAML_OUT_DIR, 'bin')
 GNUWIN32_DIR = os.path.join(WORK_DIR, 'gnuwin32')
 GNUWIN32_ZIP = 'gnuwin32.zip'
 
+options = None
 
 def IsWindows():
   return sys.platform == 'win32'
@@ -843,7 +844,7 @@ def BuildEnv(build_dir, use_gnuwin32=False, bin_subdir=False,
   return cc_env
 
 
-def LLVM(run_tests):
+def LLVM():
   buildbot.Step('LLVM')
   Mkdir(LLVM_OUT_DIR)
   cc_env = BuildEnv(LLVM_OUT_DIR, bin_subdir=True)
@@ -876,7 +877,7 @@ def LLVM(run_tests):
   proc.check_call(['ninja', 'install'] + jobs, cwd=LLVM_OUT_DIR, env=cc_env)
   CopyLLVMTools(LLVM_OUT_DIR)
 
-  if not run_tests:
+  if not options.run_tool_tests:
     return
 
   def RunWithUnixUtils(cmd, **kwargs):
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -904,9 +912,11 @@ def OCaml():
 
 def Spec():
   buildbot.Step('spec')
-  # Spec builds in-tree. Always clobber and run the tests.
+  # Spec builds in-tree. Always clobber.
   proc.check_call(['make', 'clean'], cwd=ML_DIR)
-  proc.check_call(['make', 'all'], cwd=ML_DIR)
+  proc.check_call(['make', 'opt'], cwd=ML_DIR)
```
jgravelle-google: Does `make opt` also build the spec interpreter? Afaik the waterfall uses that too.

sbc100: Yes opt == wasm == the spec interpreter.

I would put "wasm" here instead.. which might make more sense because that is what the copy command below expects.

src/build.py
```diff
@@ -749,6 +749,12 @@ def GetRepoInfo():
 
 # Build rules
 
+def RunToolTests():
+  """For local debugging it can be useful to build tools without running
+  internal tests first."""
+  return 'SKIP_TESTS' not in os.environ
```
dschuff: Thus far we have managed to avoid dependence on the environment for anything other than buildbot-specific stuff. 

Actually, thinking a little bit bigger, this is now the third "option" that modifies the build steps The others being disabling for certain OSes, and the `use_asm` option for emscripten, (which essentially has the same purpose as this one; i.e. disabling some build work when you just want to do a fast local build). We talked previously about making a more generic mechanism for the OS matrix; maybe it's time for a more general way for the build steps to take options. Not sure how ugly threading everything through the `Build` constructor would be, maybe there's even a better way. Or maybe it is just cleaner to do some global thing, I don't know. But we should probably at least ask the question.

sbc100:
OK.. you are right of course.  I did the with command line options.  PTAL

sbc100:
@dschuff @jgravelle-google either of you still at work to review this?

src/build.py
```diff
@@ -1233,15 +1241,15 @@ def Summary(repos):
       buildbot.Link('lkgr.json', cloud.Upload(info_file, lkgr_file))
 
 
-def AllBuilds(use_asm=False):
+def AllBuilds(use_asm=False, run_tests=False):
```
jfbastien: Why not just pass in options here, and have sub-tests look at run_tool_tests if they want to?

sbc100: No reason really, just following the existing convention.  We can change it later if the number of args gets out of the control.

sbc100: You might also ague it cleaner passing just want you need.

sbc100:
@dschuff  ptal

src/build.py
jfbastien: It's way more readable to use a property name instead of a bunch of boolean arguments. 🤔 
Having everything in options also prevents random option-like things from coming to life when really they should be defaulted options (instead of e.g. defaulted function parameters).

sbc100: Would you prefer a new options struct, or simply pass the command line options?

dschuff: I think +1 for an options dict instead of function args here. I think it's fine for it to match the command line options, which is probably easier than arbitrarily inventing new option names.

jfbastien: I'd just do the command line options. Where else would the values come from otherwise?

sbc100:
Refactored to make options global rather then trying to pass it everywhere.  Make this change a lot simpler.

PR #209, opened 2017-06-13
Tell CMake the version of MSVC compilers we want to use

rongjiecomputer:
In [LLVM build failure log](https://wasm-stat.us/builders/windows/builds/8807/steps/LLVM/logs/stdio/text), `vs_toolchain.py get_toolchain_dir` selected Visual Studio 2015 Update 3 (`E:\b\depot_tools\win_toolchain\vs_files\d3cb0e37bdd120ad0ac4650b674b09e81be45616`), but CMake picked up the older version of Visual Studio 2015 (`E:\b\depot_tools\win_toolchain\vs_files\95ddda401ec5678f15eeed01d2bee08fcbc5ee97\VC\bin\amd64\cl.exe`),

Workaround: explicitly setting `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER`.

`GYP_MSVS_OVERRIDE_PATH` is already set after running `vs_toolchain.py update`.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -751,7 +751,10 @@ def GetRepoInfo():
 
 def OverrideCMakeCompiler():
   if IsWindows():
-    return []
+    MSVC_COMPILER = os.path.join(os.environ['GYP_MSVS_OVERRIDE_PATH'],
+                                 'VC', 'bin', 'amd64', 'cl.exe')
+    return ['-DCMAKE_C_COMPILER=' + MSVC_COMPILER,
+            '-DCMAKE_CXX_COMPILER=' + MSVC_COMPILER]
   return ['-DCMAKE_C_COMPILER=' + CC,
           '-DCMAKE_CXX_COMPILER=' + CXX]
 
--- a/src/host_toolchains.py
+++ b/src/host_toolchains.py
@@ -25,8 +25,9 @@ WORK_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'work')
 CR_BUILD_DIR = os.path.join(WORK_DIR, 'build')
 SETUP_TOOLCHAIN = os.path.join(CR_BUILD_DIR, 'toolchain', 'win',
                                'setup_toolchain.py')
-VS_TOOLCHAIN = os.path.join(CR_BUILD_DIR, 'vs_toolchain.py')
-WIN_TOOLCHAIN_JSON = os.path.join(CR_BUILD_DIR, 'win_toolchain.json')
+V8_SRC_DIR = os.path.join(WORK_DIR, 'v8', 'v8')
+VS_TOOLCHAIN = os.path.join(V8_SRC_DIR, 'build', 'vs_toolchain.py')
+WIN_TOOLCHAIN_JSON = os.path.join(V8_SRC_DIR, 'build', 'win_toolchain.json')
 
 os.environ['GYP_MSVS_VERSION'] = '2015'
 


sbc100:
Do you know how it could be that CMake is finding one of the toolchain in depot_tools\win_toolchain but not the right one?  Somebody must be adding it to the path... seems like whereever that is being done would be where to make the fix?

rongjiecomputer:
I tried to find all log files containing "vs_files" with <kbd>Ctrl</kbd>+<kbd>F</kbd>. Only two found:

- https://wasm-stat.us/builders/windows/builds/8878/steps/Sync%20Repos/logs/stdio
- https://wasm-stat.us/builders/windows/builds/8878/steps/LLVM/logs/stdio

You can see that `vs_toolchain.py` in "Sync Repos" and "LLVM" step is copying `.dll` from `d3cb0e37bdd120ad0ac4650b674b09e81be45616` which is Visual Studio 2015 Update 3. However, in the ninja output, `cl.exe` from `95ddda401ec5678f15eeed01d2bee08fcbc5ee97` is used, but headers are from `d3cb0e37bdd120ad0ac4650b674b09e81be45616` (How can they get mixed?!)

Both versions of Visual Studio are not in`PATH`:

 `E:\b\cipd_path_tools;E:\b\cipd_client;E:\b\depot_tools\python276_bin;E:\b\depot_tools\python276_bin\Scripts;E:\b\depot_tools;E:\b\depot_tools\python276_bin;C:\Windows\system32;C:\Windows\system32\WBEM;C:\Program Files\7-Zip;C:\cmake\bin;c:\windows\system32\windowspowershell\v1.0;E:\b\build_internal\tools`

From my personal experience dealing with CMake when Visual Studio is installed in non-standard location, CMake does a very poor job in guessing which compiler I want it to use (especially when I use Ninja).

For example: Visual Studio 2017 comes in front part of `PATH` and Mingw64 GCC is in later part of `PATH`, after running `cmake -GNinja`, CMake still chooses GCC. Instead of fixing this bug, CMake developers gave me these workarounds:

- Use `-DCMAKE_IGNORE_PATH=C:\Mingw64\bin` (not applicable in this case since both Visual Studio are not in `PATH`)
- Use `-DCMAKE_C_COMPILER` and `-DCMAKE_CXX_COMPILER`
- Set environment variable `CC` and `CXX`.

sbc100:
maybe @dschuff has some insight here?

rongjiecomputer:
Ping?

src/build.py
```diff
@@ -751,7 +751,10 @@ def GetRepoInfo():
 
 def OverrideCMakeCompiler():
   if IsWindows():
-    return []
+    MSVC_COMPILER = os.path.join(os.environ['GYP_MSVS_OVERRIDE_PATH'],
```
dschuff: Is this environment variable guaranteed to exist? If so, how; and if not, then we probably don't want to make the explicit override unconditional.

rongjiecomputer: It is guaranteed to exist after running `vs_toolchain.py` (which always run before building anything).

dschuff:
@rongjiecomputer I cleared out the old build files from the windows builder (clobber build) and the problem seems to have gone away. (https://wasm-stat.us/builders/windows/builds/9093/steps/LLVM/logs/stdio) Although now it has a different problem that I can't reproduce locally. Does it behave the same for you?

rongjiecomputer:
@dschuff I have seen this in my local build. The failure is about unable to generate `Intrinsics.gen` when running `bin\llvm-tblgen.exe`. This is because everything is dynamically linked to VC++ libraries, but the required dlls are not in `bin\` folder.

Just copy `concrt140.dll`, `msvcp140.dll`, `vccorlib140.dll` and `vcruntime140.dll` to `bin\` folder.

rongjiecomputer:
From the log, `msvcp140.dll`, `vccorlib140.dll` and `vcruntime140.dll` are copied by `vs_toolchain.py` already, so only `concrt140.dll` is left to be manually copied.

dschuff:
Actually the assert is lib\support\Windows/Signals.inc, line 424, which hits when dbghelp APIs can't be loaded. Although dbghelp.dll does appear to be present? So I don't know what's up with that. Anyway, we don't need to debug that here, but I don't think we need this PR?

rongjiecomputer:
This PR is no longer needed, closing now.

PR #212, opened 2017-06-13
Add test for SIMD operations compiled to intrinsic instructions.

RebeccaRGBA:


src/build.py
```diff
@@ -1401,12 +1401,23 @@ def TestEmtestAsm2Wasm():
       warn_only=False)
 
 
+def TestWasmSimd():
+  script = os.path.join(SCRIPT_DIR, 'test_wasm_simd.py')
+  clang = Executable(os.path.join(INSTALL_BIN, 'clang'))
+  include = os.path.join(EMSCRIPTEN_SRC_DIR, 'system', 'include')
+  try:
+    proc.check_call([script, clang, include])
```
jgravelle-google: Instead of `proc.check_call`, we could `import test_wasm_simd` and call `test_wasm_simd.test_wasm_simd(clang, include)` directly

src/test_wasm_simd.py
```diff
@@ -0,0 +1,127 @@
...
+    print ('================================================================')
+    return (False, err)
+
+def test_wasm_simd_instruction(datatype, operation, arity, returntype, opcode, clang, include):
+  filename = create_test_file(datatype, operation, arity, returntype)
+  (passed, output) = compile_test_file(filename, clang, include)
+  if passed:
+    (passed, err) = check_test_output(output, opcode, arity)
+    if passed:
+      print ('%s %s: PASSED' % (datatype, operation))
```
jgravelle-google: Could probably factor out `print ('%s %s:' % (datatype, operation)),` to reduce repetition.
Could also probably write that as `print datatype, operation, ':',`

src/test_wasm_simd.py
```diff
@@ -0,0 +1,127 @@
...
+    f.write('}\n')
+  f.close()
+  return f.name
+
+def compile_test_file(filename, clang, include):
+  p = subprocess.Popen([clang, '-w', '-S', '-emit-llvm', filename,
+                        '-O3', '-o-', '-I', include],
+                       stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+  (output, err) = p.communicate()
+  if output == '':
```
sbc100: Maybe "if p.returncode != 0:" is more precise?

src/test_wasm_simd.py
```diff
@@ -0,0 +1,127 @@
...
+      print ('%s %s: PASSED' % (datatype, operation))
+    else:
+      print ('%s %s: FAILED: %s' % (datatype, operation, err))
+  else:
+    print ('%s %s: FAILED: %s' % (datatype, operation, output))
+  os.remove(filename)
+  return passed
+
+def test_wasm_simd(clang, include):
+  passed = 0
```
jgravelle-google: We probably also want to track cases that `passed_unexpectedly`, like most of the other tests do.

src/test_wasm_simd.py
```diff
@@ -0,0 +1,127 @@
...
+    f.write('  return emscripten_%s_%s(a);\n' % (datatype, operation))
+    f.write('}\n')
+  f.close()
+  return f.name
+
+def compile_test_file(filename, clang, include):
+  p = subprocess.Popen([clang, '-w', '-S', '-emit-llvm', filename,
+                        '-O3', '-o-', '-I', include],
+                       stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+  (output, err) = p.communicate()
```
sbc100: Call these "stderr, stdout" perhaps for consistency?

Also I think that the python style is to not include braces when destructuring like this.

src/test_wasm_simd.py
```diff
@@ -0,0 +1,127 @@
+#! /usr/bin/env python
```
sbc100: Maybe a docstring here giving an overview of what this script does?

src/test_wasm_simd.py
```diff
@@ -0,0 +1,127 @@
...
+  for (operation, arity, datatype, returntype, opcode, expect_failure) in wasm_simd_test_cases:
+    if test_wasm_simd_instruction(datatype, operation, arity, returntype, opcode, clang, include):
+      passed += 1
+    elif expect_failure:
+      failed_expectedly += 1
+    else:
+      failed_unexpectedly += 1
+  print ('Passes: %d' % passed)
+  print ('Failures (expected): %d' % failed_expectedly)
+  print ('Failures (unexpected): %d' % failed_unexpectedly)
```
sbc100: No space after print?  I don't write much python3.. is the space some kind of compatibility thing?  Should this script include 'from __future__ import print_function'?

src/test_wasm_simd.py
sbc100: Remove space after #!?

src/test_wasm_simd.py
```diff
@@ -0,0 +1,127 @@
...
+
+import os
+import re
+import subprocess
+import sys
+import tempfile
+
+wasm_simd_test_cases = [
+  # Oper Arity ArgType     ReturnType   Opcode   KnownFailure
+  ('add',  2, 'int32x4',   'int32x4',   'add',   False),
```
jgravelle-google: My unfiltered thoughts here are:

1. Ooh, aligned columns are nice
1. Hm, `True`/`False` are unfortunate literals to have. `True` means "should fail" in this context, which isn't necessarily intuitive. `False` also doesn't map nicely to "should fail". An enum would be nice, but, python.
1. It looks like the return types are always the same as the arg types? That probably isn't always true but happens to be for this set of ops.
1. I feel like having a data class here would make any given line more immediately readable. Something like `TestCase(operation='add', arity=2, arg_type='int32x4')`, and `TestCase(operation='min', arity=2, arg_type='float64x2', opcode='fmin', expected_fail=True)`. Then internal to that constructor we can default opcode to operation when not set, return type defaults to arg type, and expected fail defaults to False. This fixes the known failure boolean mapping.
1. Having the keyword arguments and the constructor adds line length though. If that takes things over 80 characters it makes the whole thing way less skimmable. It also makes aligning the columns meaningfully impossible, again impairing skimmability. And not having the keword args puts us right back in tuple-land.

So... Yeah I don't know what's best here. I'm ok with leaving this as-is, up to you.

src/test_wasm_simd.py
rongjiecomputer: This is python3 syntax, whitespace before `(` does not change anything. `from __future__ import print_function` should be added.

jgravelle-google: Here it's printing the string-format expression `'str with %s' % 'format'`. Strictly speaking it doesn't need parens to do so. I believe python2 syntax-wise it does need the space, although it sure *looks* a lot like python3 function printing.

Also because it's just calling print, we can drop the `%` formatting entirely and just do `print 'Passes:', passed`

jfbastien:
Which wasm VMs are intended to run these tests?

flagxor:
These are intended to vet that the toolchain's clang leaves SIMD intact.
For now, this is all in the prototype category.

RebeccaRGBA:
Fixed the style issues caught by Travis. PTAL.

src/test_wasm_simd.py
```diff
@@ -0,0 +1,150 @@
...
+from __future__ import print_function
+import os
+import re
+import subprocess
+import sys
+import tempfile
+
+wasm_simd_test_cases = [
+  # Oper Arity ArgType     ReturnType   Opcode   KnownFailure
+  ('add',  2, 'int32x4',   'int32x4',   'add',   False),  # noqa
```
jgravelle-google: `noqa` seems like a pretty big hammer, but I don't know off-hand what would be better. This is probably fine.

PR #213, opened 2017-06-20
Pass extras to outname_ctor 

sbc100:
No need the callable class.

### DIFF
--- a/src/compile_torture_tests.py
+++ b/src/compile_torture_tests.py
@@ -28,18 +28,10 @@ def c_compile(infile, outfile, extras):
   return [extras['c'], infile, '-o', outfile] + extras['cflags']
 
 
-class Outname:
-  """Create the output file's name. A local function passed to testing.execute
-  would be simpler, but it fails to pickle when the test driver calls pool.map.
-  So we manually package the suffix in a class.
-  """
-  def __init__(self, suffix):
-    self.suffix = suffix
-
-  def __call__(self, outdir, infile):
-    basename = os.path.basename(infile)
-    outname = basename + self.suffix
-    return os.path.join(outdir, outname)
+def create_outname(outdir, infile, extras):
+  basename = os.path.basename(infile)
+  outname = basename + extras['suffix']
+  return os.path.join(outdir, outname)
 
 
 def run(c, cxx, testsuite, sysroot_dir, fails, out, config):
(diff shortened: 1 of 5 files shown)


src/compile_torture_tests.py
```diff
@@ -81,9 +73,9 @@ def run(c, cxx, testsuite, sysroot_dir, fails, out, config):
   result = testing.execute(
       tester=testing.Tester(
           command_ctor=c_compile,
-          outname_ctor=Outname(suffix),
```
jgravelle-google: How does this pass `extras` to `create_outname`?
https://github.com/WebAssembly/waterfall/blob/master/src/testing.py#L78 only has it called with `outdir, infile`

sbc100: Sorry, I was trying to split this change out from anther.. my mistake.

sbc100: Fixed

src/testing.py
```diff
@@ -75,7 +75,10 @@ def setlimits():
   def __call__(self, test_file):
     """Execute a single test."""
     basename = os.path.basename(test_file)
-    outfile = self.outname_ctor(self.outdir, test_file) if self.outdir else ''
+    if self.outdir:
+      outfile = self.outname_ctor(self.outdir, test_file, self.extras)
+    else:
+      outfile = None
```
jgravelle-google: `outfile` should still be `''` instead of `None`; calling `os.path.join(None, 'foo')` throws an `AttributeError`

sbc100: I think in this case we want to signal to command_ctor that there is no outfile, passing empty string is actually worse because then someone might try to use it as a filename. 

The only user who currently cares is execute_file.py:execute, and it just check "if outfile" which works either way.  But I think empty string is wrong.

jgravelle-google: Good point, in this context not having an outdir is probably a problem. And if we really want to dump output files to the current directory, we can pass `'.'` as the outdir.
So then, that's probably better encoded as `assert self.outdir, "Need to explicitly specify the output directory"` or so, rather than relying on `os.path.join` to probably blow up later.

sbc100: Well the status quo is that execute_files seems to handle have no outfile at all.  maybe it doesn't need that anymore and we can remove support, but that would be a separate change I think

sbc100: To keep this change targetted maybe I'll switch back to ''?

jgravelle-google: Yeah, that seems right

sbc100: Done

PR #215, opened 2017-06-20
Build musl as wasm object files as well as .s files

sbc100:
The musl build script now supports build with with
the wasm32-unknown-unknown-target.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1043,7 +1043,7 @@ def Musl():
                             os.path.join(
                                 V8_SRC_DIR, 'third_party', 'cygwin', 'bin'))
 
-    # Build musl directly wasm object files in an ar library
+    # Build musl directly to wasm object files in an ar library
     proc.check_call([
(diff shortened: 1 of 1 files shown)


dschuff:
So currently the existing d8-musl build builds musl .s files, converts them to .wasm, and then passes all of them into special loader logic in wasm.js that loads all the .wasm files. That was nice as an experimental way to test that LLVM could build libc, and to test a standalone musl, but it wasn't really used for anything.  I think now is the time to just replace it wholesale with the thing you're building, which will build libc into wasm object files and link them with lld. IIUC with this PR we aren't quite yet able to make a runnable wasm binary, but we are fairly close, and it's probably not worth spending any effort to keep the .s build working.

sbc100:
Yes, my plan is to remove the old path once the lld path looks good.  I am able to run the resulting wasm files, but I wanted to incrementally land this.   So the next PR will be linking and executing.

PR #217, opened 2017-06-23
Remove FastISel expected failures

jgravelle-google:
LLVM change http://llvm.org/viewvc/llvm-project?view=revision&revision=306060 fixed these FastISel regressions


sbc100:
Nice!

PR #218, opened 2017-06-30
Run jsc tests with --fast

jfbastien:
For some reason the waterfall seems to dislike the wasm-no-cjit-yes-tls-context tests. They work on the JSC infra, and I'm not sure it's useful to run all tests 5-ways for the waterfall anyways. --fast will only run with the basic command-line setup, not the extra 4 variants we otherwise have.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -884,7 +884,7 @@ def Jsc():
   proc.check_call(['../Tools/Scripts/run-javascriptcore-tests',
                    '--root=bin',
                    '--filter', 'wasm',
-                   '--no-build', '--no-testapi'],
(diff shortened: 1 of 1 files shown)


dschuff:
Thanks!

dschuff:
BTW this actually also reproduces on my laptop, which is a more conventional setup than the bot infrastructure.

jfbastien:
Weird... I'll take a look at some point. Seems to work fine for us. What's the backtrace?

dschuff:
Hm, still seems to have the same failures: https://wasm-stat.us/builders/mac/builds/6377

PR #220, opened 2017-07-11
Update lld git url to github.com/WebAssembly/lld

sbc100:


src/build.py
```diff
@@ -113,7 +113,7 @@
 WEBKIT_GIT_BASE = 'https://github.com/WebKit/'
 
 # TODO(sbc): Remove this once lld changes are upstream
-LLD_GIT_BASE = 'https://github.com/sbc100/'
+LLD_GIT_BASE = 'https://github.com/WebAssembly/'
 LLD_BRANCH = 'add_wasm_linker'
```
dschuff: Actually we need to change this branch name now too

sbc100: OK

PR #222, opened 2017-07-12
Bring back gnuwin32 tools for preparing musl sources on Windows

dschuff:
The copy of cygwin has been removed from V8, so we need something
to run bash and sed for generating some musl headers. The 'git bash'
hack we use for LLVM tests stil doesn't work for this, so bring back
our package of gnuwin32 tools. This essentially reverts rev 2f160b0
and uses it for musl instead of LLVM.


sbc100:
Hmm I thought musl was bulld with just a python script.

dschuff:
It's built with a python script that runs `sed` and `sh` :grin:

PR #223, opened 2017-07-13
Make JSC build failure not abort the whole process

aheejin:
Currently JSC build is using cmake build, which is not the official
build, and this makes JSC build flaky. This makes JSC build failure
not stop the whole waterfall process so that we can proceed to the
test stage.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -910,7 +910,7 @@ def Jsc():
   except proc.CalledProcessError:
     # JSC cmake build is flaky because it is not the official build. For the
     # moment make this not abort the whole process.
-    buildbot.Fail(True)
+    buildbot.Warn()
 
 
 def Wabt():
@@ -1219,7 +1219,10 @@ def ExecuteLLVMTorture(name, runner, indir, fails, attributes, extension,
       wasmjs=wasmjs,
       extra_files=extra_files)
   if 0 != unexpected_result_count:
-      buildbot.Fail(warn_only)
+      if warn_only:
+        buildbot.Warn()
+      else:
+        buildbot.Fail()
   return outdir
 
 
@@ -1469,7 +1472,10 @@ def ExecuteEmscriptenTestSuite(name, config, outdir, warn_only):
          'binaryen2', '--em-config', config],
         cwd=outdir)
   except proc.CalledProcessError:
-    buildbot.Fail(warn_only)
+    if warn_only:
+      buildbot.Warn()
+    else:
+      buildbot.Fail()
 
 
 def TestEmtest():
--- a/src/build.py
+++ b/src/build.py
@@ -1406,7 +1406,7 @@ def TestBare():
       wasmjs=os.path.join(INSTALL_LIB, 'wasm.js'),
       extra_files=[os.path.join(INSTALL_LIB, 'musl.wasm')])
 
-  if IsMac() and 'JSC' not in buildbot.WarnedList():
+  if IsMac() and not buildbot.DidStepFailOrWarn('JSC'):
     ExecuteLLVMTorture(
         name='jsc',
         runner=os.path.join(INSTALL_BIN, 'jsc'),
--- a/src/build.py
+++ b/src/build.py
@@ -1264,6 +1264,9 @@ def Summary(repos):
   print 'Failed steps: %s.' % buildbot.Failed()
   for step in buildbot.FailedList():
     print '    %s' % step
+  print 'Warned steps: %s.' % buildbot.Warned()
+  for step in buildbot.WarnedList():
+    print '    %s' % step
 
   if IsBuildbot():
     latest_file = '%s/%s' % (BUILDBOT_BUILDERNAME, 'latest.json')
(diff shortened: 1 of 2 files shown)


jgravelle-google:
What happens when we try to run the JSC tests during the test phase, with no jsc built? Last time I tried it threw an assert, which stopped the build early as well.

src/build.py
```diff
@@ -1258,6 +1264,9 @@ def Summary(repos):
   print 'Failed steps: %s.' % buildbot.Failed()
   for step in buildbot.FailedList():
     print '    %s' % step
+  print 'Warned steps: %s.' % buildbot.Warned()
+  for step in buildbot.WarnedList():
+    print '    %s' % step
 
   if IsBuildbot():
     latest_file = '%s/%s' % (BUILDBOT_BUILDERNAME, 'latest.json')
```
dschuff: We should also maybe not update LKGR if there were warned steps (see L1275 below). I seemed to recall @jfbastien wanting that on #177. It's a little ambiguous about what LKGR actually means, especially since the different OS bots test some different things, but it seems fine to require that JSC works for now.

src/buildbot.py
```diff
@@ -38,7 +39,7 @@ def Fail(warn_only=False):
   """Mark one step as failing, but keep going."""
   sys.stdout.flush()
   if warn_only:
-    sys.stdout.write('\n@@@STEP_WARNINGS@@@\n')
+    Warn()
```
dschuff: It's a little bit nonobvious what `Warn()` means (i.e. print a warning) vs `Fail()` with a boolean flag.  I think it would be more clear if we kept `Fail()` just for hard failures, and then have `Warn()` do what `Fail(warn_only=True)` does today. Then they would be completely symmetric. 

Or,

dschuff: Oh, I didn't finish. The other option might be to keep the boolean flags, and put the printing together in one function with a boolean switch for the type.

aheejin: 1. I think it would be more clear if we kept Fail() just for hard failures, and then have Warn() do what Fail(warn_only=True) does today.
--> I can do that

2. The other option might be to keep the boolean flags, and put the printing together in one function with a boolean switch for the type.
--> The reason I made Warn() is I wanted buildbot.WarnedList() (to be used for condition whether to test JSC tests), and it would look weird if they don't have symmetric methods for Fail() and Failed(). If we delete Warn() and move the printing to Fail(), the symmetry would break, but I don't mind too much either.

src/build.py
```diff
@@ -1397,7 +1406,7 @@ def TestBare():
       wasmjs=os.path.join(INSTALL_LIB, 'wasm.js'),
       extra_files=[os.path.join(INSTALL_LIB, 'musl.wasm')])
 
-  if IsMac():
+  if IsMac() and 'JSC' not in buildbot.WarnedList():
```
dschuff: This is a step in the right direction, and we should keep it for this PR. But we might want to do the same thing for other tests, e.g. the bare tests depend on the musl build, and the emscripten tests depend on the emscripten build. So in the future we might consider some more generic dpendence mechanism.

jgravelle-google: It feels kind of unfortunate to me that this line has to care whether JSC failing is a `Fail` or a `Warn`. Maybe `and not buildbot.DidStepFail('JSC')`, where `DidStepFail` checks the warning and failure arrays?

src/build.py
jgravelle-google: That sounds kinda iffy, because then we'd have Green builds not updating LKGR. We could introduce Yellow builds, but then why not just have Red builds?

Alternatively: we make JSC failures errors, but let tests run when there are errors? Although that depends on the broader dependency-handling mechanism.

Alternatively: we hack/hardcode this to not upload LKGR if JSC failed (while turning the bot red here?), with a TODO to make that more general.

src/build.py
dschuff: Yeah, I like that better. I guess that might go along with keeping `Fail()` as-is, and `DidStepFail` doesn't care whether it warned or failed.

aheejin: I added `DidStepFailOrWarn()`, because `DidStepFail` sounds like it might not include warnings.

src/build.py
dschuff: Yeah, it kind of really boils down to who cares about what. The test-dependency thing is solvable the same way regardless. I think this situation came down to, "we care less about JSC and also about LKGR" (i.e. we don't really use it for anything right now) and "JF cares more about JSC and also more about LKGR" (presumably he wouldn't have asked for that behavior if he didn't care). My opinion for now is that if nobody cares enough about JSC to keep it up to date (and this seems empirically true at least right now), then I don't think it should turn the bot red. But I also don't care about LKGR all that much right now either, so I'm happy to defer to what JF wants on that.

src/build.py
```diff
@@ -910,7 +910,7 @@ def Jsc():
   except proc.CalledProcessError:
     # JSC cmake build is flaky because it is not the official build. For the
     # moment make this not abort the whole process.
-    buildbot.Fail(True)
+    buildbot.Warn()
```
jgravelle-google: Nice

src/buildbot.py
```diff
@@ -55,6 +52,8 @@ def FailedList():
 
 
 def Warn():
+  """We mark this step as failing, but this step is flaky so we don't care
```
jgravelle-google: Yay docstrings

aheejin:
Nevermind the revert commit; it was a click mistake :(

PR #225, opened 2017-08-21
Update GCC test failure descriptions

aheejin:
This PR will be used to update GCC test failure descriptions to track
failures we need to fix. Related to kripken/emscripten#5488.


aheejin:
Thanks, I'll probably leave this open to make more commits.

PR #228, opened 2017-08-23
Update lld expectations now that we always write an empty table

sbc100:
See:
https://github.com/WebAssembly/lld/commit/b5e5931adccf073e4af2a0df968cfaff4675333a


dschuff:
Although it probably makes more sense to have an lld flag to control table emission and make clang always emit that (as opposed to just always emitting a table section).

PR #229, opened 2017-08-24
Make waterfall test multiple optimization levels

aheejin:
Currently bare tests are only compiled with -O0, whereas emcc/binaryen
tests are only compiled with -O2. This makes bare tests compiled with
both -O0 and -O2, and emcc/binaryen tests with both -O0 and -O3.
(The reason for not -O2 again but -O3 is people tend to use -O3 a lot in
real use cases and test emcc with a different set of flags will increase
test coverage.)

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -195,6 +195,10 @@ BINARYEN_SHELL_KNOWN_TORTURE_FAILURES = [
     os.path.join(BINARYEN_SRC_DIR, 'test',
                  's2wasm_known_binaryen_shell_test_failures.txt')]
 
+# Optimization levels
+BARE_TEST_OPT_FLAGS = ['O0', 'O2']
+BINARYEN_TEST_OPT_FLAGS = ['O0', 'O3']
+
 
 NPROC = multiprocessing.cpu_count()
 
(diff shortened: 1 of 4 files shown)


src/build.py
```diff
@@ -195,6 +195,10 @@ def NodePlatformName():
     os.path.join(BINARYEN_SRC_DIR, 'test',
                  's2wasm_known_binaryen_shell_test_failures.txt')]
 
+# Optimization levels
+BARE_TEST_OPT_FLAGS = ['O0', 'O2']
+BINARYEN_TEST_OPT_FLAGS = ['O0', 'O3']
```
dschuff: Let's call this `EMSCRIPTEN_TEST_OPT_FLAGS` since the bare tests also use pieces of Binaryen. (And Binaryen is really just an implementation detail of emscripten).

src/build.py
```diff
@@ -1130,99 +1134,112 @@ def DebianPackage():
...
-  unexpected_result_count = compile_torture_tests.run(
-      c=c, cxx=cxx, testsuite=GCC_TEST_DIR,
-      sysroot_dir=INSTALL_SYSROOT,
-      fails=LLVM_KNOWN_TORTURE_FAILURES,
-      out=outdir,
-      config='wasm-' + extension)
-  UploadArchive('torture-' + extension, Archive(outdir))
-  if 0 != unexpected_result_count:
-    buildbot.Fail()
+  for opt in BARE_TEST_OPT_FLAGS:
```
dschuff: I think it might be a little more clear/readable if CompileLLVMTorture (and Link/Assemble/Execute) just did a single compile (and took an argument for the opt level), and we put the `for opt in BARE_TEST_OPT_FLAGS` out in the caller. 
That way we'd get one test run per call, and that would be consistent with the way we call it once for lld and once for s2wasm. And if we later decide we don't want the full cross-product of optlevel x lld it would be easier to change.

dschuff: Although I guess the downside would be that we'd have 3 sets of `for opt in flags: LinkLLVMTorture()` if we wanted to keep the same test ordering that we have here, which I think we do. Maybe that's not so bad though.

src/build.py
```diff
@@ -1130,99 +1134,112 @@ def DebianPackage():
...
+    Remove(outsubdir)
+    Mkdir(outsubdir)
+    input_pattern = os.path.join(indir, opt, '*.' + extension)
+    unexpected_result_count = link_assembly_files.run(
+        linker=linker, files=input_pattern, fails=fails, attributes=[opt],
+        out=outsubdir, args=args)
+    UploadArchive('torture-%s-%s' % (name, opt), Archive(outsubdir))
+    if 0 != unexpected_result_count:
+      buildbot.Fail()
+  return os.path.join(WORK_DIR, 'torture-%s' % name)
```
jgravelle-google: Nit: Instead of repeating `os.path.join(WORK_DIR, 'torture-%s' % name)`, we can create a temporary `out` and `os.path.join(out, opt)`, which is slightly-better-documenting what that expression means (and similarly explains how `outsubdir` is a subdir)

src/compile_torture_tests.py
```diff
@@ -68,7 +68,7 @@ def run(c, cxx, testsuite, sysroot_dir, fails, out, config):
                                     c_torture)
   assert os.path.isdir(out), 'Cannot find outdir %s' % out
   c_test_files = glob.glob(os.path.join(c_torture, '*c'))
-  cflags = cflags_common + cflags_extra[config]
+  cflags = cflags_common + cflags_extra[config] + ['-' + opt]
```
jgravelle-google: should `['-' + opt]` be added to `cflags_common`?

src/build.py
```diff
@@ -1130,99 +1134,112 @@ def DebianPackage():
...
-    buildbot.Fail()
-  return out
+def LinkLLVMTorture(name, linker, fails, indir, extension, args=None):
+  for opt in BARE_TEST_OPT_FLAGS:
+    buildbot.Step('Link LLVM Torture (%s, %s)' % (name, opt))
+    assert os.path.isfile(linker), 'Cannot find linker at %s' % linker
+    outsubdir = os.path.join(WORK_DIR, 'torture-%s' % name, opt)
+    Remove(outsubdir)
+    Mkdir(outsubdir)
+    input_pattern = os.path.join(indir, opt, '*.' + extension)
```
jgravelle-google: Nice, I think that simplifies this function's callsites

src/build.py
aheejin: Done. Actually I was torn between the two as well and chose `BINARYEN` to be consistent with the function name `CompileLLVMTortureBinaryen`.

src/compile_torture_tests.py
aheejin: Good point

src/build.py
aheejin: Done

aheejin: Hmm actually if I change the code according to this (https://github.com/WebAssembly/waterfall/pull/229#discussion_r135059826), this subdir thing wouldn't even be necessary, because I have to supply outsubdir directly as an argument of **Torture() functions...

src/build.py
jgravelle-google: Er, should probably do this first

aheejin: Done

src/build.py
```diff
@@ -1323,144 +1327,199 @@ def Test(self):
...
   # Compile
-  CompileLLVMTorture('s', TORTURE_S_OUT_DIR)
-  CompileLLVMTorture('o', TORTURE_O_OUT_DIR)
+  for opt in BARE_TEST_OPT_FLAGS:
+    CompileLLVMTorture('s', os.path.join(TORTURE_S_OUT_DIR, opt), opt)
+  for opt in BARE_TEST_OPT_FLAGS:
+    CompileLLVMTorture('o', os.path.join(TORTURE_O_OUT_DIR, opt), opt)
 
   # Link/Assemble
+  lld_musl_out = {}
```
jgravelle-google: Hm, this feels squirrely to me, but I think I'm just being weird.

This is a very natural extension of precomputing the directory name. I much prefer it to reusing the `os.path.join` logic.

I think my unease comes from how this introduces state that gets computed as the program runs, where we
1. compute and store -O0 dir
1. link -O0 files
1. compute and store -O2 dir
1. link -O2 files
1. run -O0 files
1. run -O2 files
and that interleaving of storing the dir with linking seems off. In theory we could add another `for opt in BARE_TEST_OPT_FLAGS` to precompute the dirs, then extract that into a function giving us something like `lld_musl_out = PrecomputeOutDirs(os.path.join(WORK_DIR, 'torture-lld-musl'), BARE_TEST_OPT_FLAGS)`

I also like that because it avoids repeating this pattern for the other cases, but, again, it might be total overkill.

aheejin: Yeah, I wasn't super satisfied with the flow either actually. So the original flow had this kind of chain:
```
s2wasm_out = LinkLLVMTorture('s2wasm', ...)
wast2wasm_out = AssembleLLVMTorture(..., indir=s2wasm_out, ...)
ExecuteLLVMTorture(..., indir=wast2wasm_out, ...)
```

But as we introduce a compilation flag loop for each of them, this chain does not work anymore, which was the reason I introduced this map thing, which wasn't super tidy either.

And the other thing is, all those ***LLVMTorture functions had different ways of getting `indir` and `outdir` - some of them compute them within the function while others take them as arguments. Some function only take indir as an argument and others take both `indir`/`outdir`, and all of this was confusing, so I made all of torture functions take both `indir`/`outdir` as arguments (except for `CompileLLVMTorture` and `CompileLLVMTortureBinaryen`, which takes c codes form gcc test suite directory). Then I couldn't help but wonder, why all these functions take `outdir` as their parameter and return the very same thing as a result? So I decided to delete all the return values from all torture functions and compute all directories only from outside.

So far was the not-very-interesting flow of my thoughts. Anyway I created `GetTortureDir` function and made all other functions inquire this function to get torture directories. PTAL.

src/build.py
```diff
@@ -1325,57 +1325,61 @@ def Test(self):
     self.runnable()
 
 
+def GetTortureDir(name, opt):
+  dirs = {
```
jgravelle-google: This is cleverly done, it makes the special cases explicit while cutting down on caller boilerplate and covering all the cases. Nice

PR #237, opened 2017-09-12
Add icudtl.dat to V8 install

jgravelle-google:
V8 commit d9a25842d3aba82ead5e06c7fd7cb4896e6ad202 changed Date objects
to use ICU timezone data, which is read from the icudtl.dat file.


dschuff:
Does that data file just include timezone data or other stuff too?
But yeah, you could leave a comment like "include the v8 blobs and the ICU data file for timezone data"

PR #239, opened 2017-09-21
Split LLVM regression into its own buildbot step

dschuff:
The regression test runner is currently failing on windows for obscure reasons, but the rest of the tests seem to work, so we can at least run them rather than bailing out completely.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1190,7 +1190,7 @@ def LinkLLVMTorture(name, linker, fails, indir, outdir, extension,
       out=outdir, args=args)
   UploadArchive('torture-%s-%s' % (name, opt), Archive(outdir))
   if 0 != unexpected_result_count:
-    buildbot.FailUnless(lambda: IsWindows() and 'lld' in name)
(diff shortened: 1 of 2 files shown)


src/build.py
```diff
@@ -1529,10 +1530,7 @@ def ExecuteEmscriptenTestSuite(name, config, outdir, warn_only):
          'binaryen2', '--em-config', config],
         cwd=outdir)
   except proc.CalledProcessError:
-    if warn_only:
-      buildbot.Warn()
-    else:
-      buildbot.Fail()
+    buildbot.FailUnless(lambda: warn_only)
```
sbc100: Why bother with these lambdas?  Why not just pass the variable?

dschuff: My original iteration of this PR had a call that was like
`buildbot.FailUnless(lambda: IsWindows() and 'lld' in name)` i.e. a more complex condition. It made that use case much nicer at a pretty small complexity cost.

PR #240, opened 2017-09-22
Reduce torture test STACK_SIZE define

dschuff:
A stack overflow in 960521-1.c results in an infinite loop and hangs the
bots. Reducing the size used by some of the tests works around that.

### DIFF
--- a/src/compile_torture_tests.py
+++ b/src/compile_torture_tests.py
@@ -36,7 +36,7 @@ def create_outname(outdir, infile, extras):
 
 def run(c, cxx, testsuite, sysroot_dir, fails, out, config, opt):
   """Compile all torture tests."""
-  cflags_common = ['--std=gnu89', '-DSTACK_SIZE=1044480',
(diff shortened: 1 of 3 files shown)


aheejin:
I wonder why it results in an infinite loop?

dschuff:
I dunno, it is kind of odd. It is still in the "untriaged" section of the failures, so... I guess someone can look at it sometime? :) In one sense stack overflows aren't that interesting because it's a resource constraint that the compiler doesn't control, but we could maybe try to make sure we aren't doing anything pathological? Also if we start to have similar issues on optimized code then we should be more concerned.

aheejin:
Yeah what makes me wonder is why not _increasing_ but _decreasing_ stack size removes the stack overflow... intuitively it sounds like it should be the other way around, so :woman_shrugging:

dschuff:
Well the define that I changed reduces the stack size that the torture program thinks it has available to it. I didn't change the stack space that the linker allocates. So that part makes sense.

PR #243, opened 2017-10-05
Force clobbering on windows

aheejin:
`shutil.rmtree()` does not work when there are read-only files in a
directory. This works around that problem.

### DIFF
--- a/src/file_util.py
+++ b/src/file_util.py
@@ -50,17 +50,21 @@ def Mkdir(path):
 
 def Remove(path):
   """Remove file or directory if it exists, do nothing otherwise."""
-  if os.path.exists(path):
-    print 'Removing %s' % path
+  if not os.path.exists(path):
+    return
+  print 'Removing %s' % path
+  if sys.platform == 'win32':
+    # shutil.rmtree() may not work in Windows if a directory contains read-only
+    # files.
+    if os.path.isdir(path):
+      proc.check_call(['rmdir', '/S', '/Q', '"' + path + '"'])
+    else:
+      os.remove(path)
+  else:
     if os.path.isdir(path):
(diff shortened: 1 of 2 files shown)


src/file_util.py
```diff
@@ -47,8 +50,17 @@ def Mkdir(path):
...
+  if not os.path.exists(path):
+    return
+  print 'Removing %s' % path
+  if sys.platform == 'win32':
+    # shutil.rmtree() may not work in Windows if a directory contains read-only
+    # files.
+    if os.path.isdir(path):
+      proc.check_call(['rmdir', '/S', '/Q', '"' + path + '"'])
+    else:
+      os.remove(path)
```
jgravelle-google: We've got two nested bools, and two calls to `os.remove`, we can take both of those out by doing:
```python
if not os.path.isdir(path):
  os.remove(path)
elif sys.platform == 'win32':
  # check_call 'rmdir'
else:
  shutil.rmtree(path)
```

Now, I'm not totally convinced that that's better: it kind of interleaves the two bools and makes it a little more contextual (the `elif` selects a different condition than the `if`, which is a little odd). So, that's a non-binding suggestion, whichever you prefer.

aheejin: Then how about
```
if not os.path.isdir(path):
  os.remove(path)
  return
if sys.platform == 'win32':
  # check_call 'rmdir'
else:
  shutil.rmtree(path)
```

jgravelle-google: Oo I like it, sgtm

aheejin:
By the way it hasn't been tested because I don't have a windows machine. I know that works for windows though. May I just land it and try clobbering the windows bot?

PR #244, opened 2017-10-05
Update OCaml build to work with 4.05

dschuff:
1. Switch from a source tarball to use a release tag on the official github repo, using our standard git checkout mechanism.
2. Check out and build `ocamlbuild` which is no longer packaged with the compiler/runtime sources.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -944,8 +944,9 @@ def OCaml():
   makefile = os.path.join(OCAML_DIR, 'config', 'Makefile')
   if not os.path.isfile(makefile):
     configure = os.path.join(OCAML_DIR, 'configure')
-    cc_flag = ['-cc', CC,
-               '-aspp', CC + ' -c'] if sys.platform != 'darwin' else []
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -940,17 +940,27 @@ def Wabt():
 
 def OCaml():
   buildbot.Step('OCaml')
+  # Build the ocaml compiler and runtime
   makefile = os.path.join(OCAML_DIR, 'config', 'Makefile')
   if not os.path.isfile(makefile):
     configure = os.path.join(OCAML_DIR, 'configure')
-    cc_flag = ['-cc', CC] if sys.platform != 'darwin' else []
+    cc_flag = ['-cc', CC,
```
jgravelle-google:
```python
cc_flag = # stuff
if sys.platform != 'darwin':
  cc_flag = []
```
?

dschuff: What, you don't like 2 line one-liners? :D

PR #245, opened 2017-10-05
Fix a bug in windows clobbering

aheejin:
Shell commands like `rmdir` should have `shell=True` in `subprocess`
commands. Because `proc.check_call` assumes it takes a list and here we
want a command string, we just use `subprocess.check_call` directly here
with `shell=True`.

### DIFF
--- a/src/file_util.py
+++ b/src/file_util.py
@@ -20,8 +20,9 @@
 import errno
 import os
 import shutil
-import subprocess
 import sys
+
+import proc
 
 
 def Chdir(path):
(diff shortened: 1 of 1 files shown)


dschuff:
`proc.check_call` should pass through any kwargs directly to `subprocess.check_call`. Did you try just adding `shell=True` to `proc.check_call`?

dschuff:
I just checked on my windows box and confirmed: you should be able to just switch to a string and add `shell=True` to `proc.check_call` and it should work.

aheejin:
Done

PR #246, opened 2017-10-05
Update lld expectations temporarily

aheejin:
Update lld expectations temporarily to make the waterfall green again,
until @sbc100 is back from vacation.


aheejin:
@sbc100 Please take a look after you come back :D

PR #247, opened 2017-10-05
Remove stray spaces in subprocess messages (NFC)

aheejin:
When the `cmd` argument to `proc.check_call` is a string, the code
tried to join strings with ' ', resulting strings like `r m d i r`.


aheejin:
Done

PR #249, opened 2017-10-06
Bugfix in proc.check_call introduced by #247

aheejin:
- I accidentally deleted a line in #247. :( Restored it.
- Cosmetic change in `file_util.py` (NFC)

### DIFF
--- a/src/file_util.py
+++ b/src/file_util.py
@@ -59,7 +59,7 @@ def Remove(path):
   if sys.platform == 'win32':
     # shutil.rmtree() may not work in Windows if a directory contains read-only
     # files.
-    proc.check_call(['rmdir', '/S', '/Q', path], shell=True)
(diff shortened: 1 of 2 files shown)


aheejin:
Current mac and windows waterfall redness is because of this. Sorry 😢

src/file_util.py
```diff
@@ -59,7 +59,7 @@ def Remove(path):
   if sys.platform == 'win32':
     # shutil.rmtree() may not work in Windows if a directory contains read-only
     # files.
-    proc.check_call('rmdir /S /Q "%s"' % path, shell=True)
+    proc.check_call(['rmdir', '/S', '/Q', path], shell=True)
```
dschuff: I think `shell=True` doesn't really work with the list form for check_call. I think we just want the `SpecialCases` part.

aheejin: Done

PR #252, opened 2017-10-11
Add test binaries to JSC build

dschuff:
These new targets are needed to run the JSC tests.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -908,7 +908,8 @@ def Jsc():
     command.extend(host_toolchains.CmakeLauncherFlags())
 
     proc.check_call(command, cwd=JSC_OUT_DIR)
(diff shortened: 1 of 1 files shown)


jfbastien:
You could do this instead when testing: `--no-testapi --no-testmasm --no-testb3 --no-testair`

dschuff:
Do those tests have anything to do with wasm? If not, maybe I will just disable them.

jfbastien:
Not really. They test lower-level things, some of which are used in wasm. At the moment all relevant wasm testing for the jsc shell is done by JSTests/wasm which is all hosted within JS.

There are also Web tests in LayoutTests, but those require more than just the shell.

And I'll add more later, but no need to worry about them now!

PR #253, opened 2017-10-17
Write and check stamp file when syncing archives

sbc100:
Also, rename cmake directly to cmake-3.4.3 to more clearly
match the upstream version.

This change is in prepration for updating the cmake version.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -77,9 +77,6 @@ PREBUILT_CLANG_BIN = os.path.join(
 CC = os.path.join(PREBUILT_CLANG_BIN, 'clang')
 CXX = os.path.join(PREBUILT_CLANG_BIN, 'clang++')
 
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -481,17 +497,26 @@ def SyncToolchain(name, src_dir, git_repo):
...
+  is expected to be 'out_dir', so if 'out_dir' already exists then download
+  will be skipped.
   """
+
+  stamp_file = os.path.join(out_dir, 'stamp.txt')
   if os.path.isdir(out_dir):
-    print '%s directory already exists' % name
-    return
-  print 'Downloading %s %s from %s' % (name, version, url)
+    if os.path.exists(stamp_file):
```
dschuff: Maybe do `os.path.isfile` just to be paranoid?

PR #254, opened 2017-10-18
Fix invocation of v8 run-tests.py

sbc100:
This auto detect feature of run-tests.py was removed
upstream in:
 https://chromium-review.googlesource.com/721659


dschuff:
LGTM

PR #256, opened 2017-10-20
Add 930529-1.c at -O2 to known failures

sbc100:


jgravelle-google:
The patch I had staged looked like:
```
# Clang optimizing loop into infinite loop for -O2
930529-1.c.s.wast.wasm O2
930529-1.c.o.wasm O2
930529-1.c.js emwasm,O3 # asm2wasm uses different clang
```

Which is to say it would be nice to have a bit more detail, to merge them into the same root cause, and to also disable the emwasm -O3 version while we're at it.

sbc100:
Want to just upload yours?

jgravelle-google:
sgtm

jgravelle-google:
https://github.com/WebAssembly/waterfall/pull/257

PR #258, opened 2017-10-21
Build wasm32-unknown-unknown-wasm version of musl using its own makefile

sbc100:


jfbastien:
That's the musl that's hosted in my repo? It seems like you're making it a bit more capable and useful on its own. Do you think we should have a more official location than that? Happy to move it if so, and then I assume we can try to get it upstreamed?

sbc100:
Yes, its the musl from your repo.  I think it would be great to upstream some/all of these changes.  My understanding is that the musl developers were not interested in patches to increase portability to target new OSes?

Regarding moving the musl tree into a more official location I think that is probably a good idea.  Perhaps one day we could have this be shared with emscripten.

jfbastien:
I would rather have a free-standing thing, if its purpose is to support an Emscripten-less build.

sbc100:
What do you mean by freestanding?   (Perhaps you thought I was suggesting merging into emscripten?  I didn't mean this, I was suggesting that perhaps emscripten could one say switch to imported musl rather than containing its own fork... same for compiler-rt)

sbc100:
I'm happy to move musl repo, but that seems unrelated to this particular change so no need to block on that.

jfbastien:
Yeah no worries, I was just asking while we're here :)

dschuff:
Does this have any chance of working on Windows? If not, what are we going to do?

sbc100:
I think so.. it should only depend on 'make' plus the toolchain binaries.   So the only questions is whether 'make' is already installed?

dschuff:
We love unix utils on windows so much, we have them twice. I originally packaged the gnuwin32 tools (https://github.com/WebAssembly/waterfall/blob/master/src/build.py#L836) to run the LLVM regression tests (that's the officially recommended way) and also to run the existing libc.py (because it uses sed from musl's build system under the hood; so it's not just make). But then that stopped working for LLVM for some reason. So I tried another hack; using depot_tools 'git bash' (https://github.com/WebAssembly/waterfall/blob/master/src/build.py#L879) which has msys under the hood. But I never could get them both working for both use cases simultaneously. So currently we have both. But the LLVM reguression suite is currently in the state where it actually does successfully run under git bash (I verified this locally) but then the stdout output doesn't get piped back to the buildbot, and the exit status doesn't propagate either, which is why it's currently XFAILed. Every month during my sheriff week I poke at it a little more when I'm feeling masochistic.

sbc100:
Do you know if any of these installations have 'make'?

dschuff:
Hm, it doesn't look like either one of them does. It looks like gnuwin32 has a build of make, although I'd be pretty shocked if there weren't filename/path issues in the makefiles themselves. We could try it, or maybe msys2. Or maybe just go with the python file? Not sure how complex the build is or how hard to reproduce.

sbc100:
Closing this for now in favor of improving libc.py

PR #260, opened 2017-10-24
Pin flake8 to specific version on travis

sbc100:


sbc100:
I picked the version of flake8 that ships with trusty since that is what a lot of us have on our machines still.

sbc100:
Gonna land this one TBR, as its blocking other changes and seems pretty trivial.

PR #262, opened 2017-10-25
Add expected failure for another test with zero sized symbol

sbc100:
These should be fixed once we land https://reviews.llvm.org/D39107


aheejin:
I don't know much about the context of https://reviews.llvm.org/D39107, but if you are going to land that soon, do we need this?

sbc100:
I want to get the waterfall green first ideally.

sbc100:
Since its been while since we had a green build

dschuff:
Except the compiler_rt build is still broken, so nothing beyond that is
even building, no? Do we need a newer version of CMake?

PR #263, opened 2017-10-25
Temporarily disable compiler-rt

sbc100:


dschuff:
LGTM. is this a problem with the upstream CMake files?

sbc100:
Doh! I think it was just that I only clobbered one of the bots.. but I needed to clobber all of them

PR #268, opened 2017-10-27
Remove non-musl 'lld' configuration

sbc100:
It wasn't adding much value and maintaining it in parallel
adds a fair amount of overhead/churn.


dschuff:
cool, so now 'lld' is what 'lld-musl' was before?

sbc100:
Should I do the same for the non-musl s2wasm path too?

dschuff:
That's fine with me. I think lld is far enough along that we don't need to maintain another thing just to get the codegen test coverage.

PR #271, opened 2017-11-02
Remove tests that use musl.wasm

sbc100:
wasm.js isn't correctly linking musl.wasm so the
results are currently identical to the non-musl tests.

See https://github.com/jfbastien/musl/pull/17

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1485,17 +1485,6 @@ def TestBare():
         extension='wasm',
         opt=opt,
         wasmjs=os.path.join(INSTALL_LIB, 'wasm.js'))
-  for opt in BARE_TEST_OPT_FLAGS:
-    ExecuteLLVMTorture(
-        name='d8-musl',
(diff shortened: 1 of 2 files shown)


sbc100:
@jfbastien  what do you think of this way forward:  Fix the wasm.js so that shared library loading can (in thoery) work again, but at the same time remove the configuration here since maintaining it is not worth the effort (For example: we'd need build all the exectuables twice, once for the non-musl case and once for the musl case.  We'd also need to figure out how to put the static data form musl and the main exe in non-overlapping regions).

jfbastien:
That's fine with me.

> We'd also need to figure out how to put the static data form musl and the main exe in non-overlapping regions

Don't you already need to do this for dynamic linking in general?

sbc100:
Yes, although I think wasm.js in musl is a ways away from implementing the full dynamic linking as specified here:
https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md.   We could increase support there, but I'm not actively working on that right now.

PR #274, opened 2017-11-03
Add ability to run build.py with a limited set of torture tests

sbc100:
Also remove saving/restoring PATH in Musl() step.  This
step doesn't modify PATH

### DIFF
--- a/src/compile_torture_tests.py
+++ b/src/compile_torture_tests.py
@@ -22,6 +22,10 @@ import sys
 
 import testing
 
+# For debugging purposes set this to a source file name to test just a single
+# file.
+TEST_FILTER = None
+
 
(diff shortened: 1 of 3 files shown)


src/compile_torture_tests.py
```diff
@@ -22,6 +22,10 @@
 
 import testing
 
+# For debugging purposes set this to a source file name to test just a single
+# file.
+TEST_FILTER = None
```
dschuff: I'm ok with this, but another option would be to pass this as an argument into `c_compile` and then put `TEST_FILTER` at the top of build.py. Obviously it's more plumbing but it would mean that all the places you would randomly edit would be in build.py. Either way.

PR #277, opened 2017-11-07
Use clang driver when linking with lld

sbc100:
This gives us libc and compiler-rt on the link line for free
and excercises the clang driver.


dschuff:
sweet :+1: 
Maybe we should make the 'wasm' target the default (instead of elf) and make a wasm32-unknown-unknown-clang, and then we wouldn't even need the --target flag.

PR #281, opened 2017-11-07
Use musl-provided wasm.syms

sbc100:
No need to auto-generate it anymore.
See: https://github.com/jfbastien/musl/pull/21


dschuff:
LGTM

PR #285, opened 2017-11-10
Replace compiler-rt CMake flags with a CMake toolchain description file

dschuff:
This prevents CMake from using any host system headers when compiling, (this just happens to work on the linux bot but it breaks the mac bot).


dschuff:
For now I haven't brought back the native compiler-rt build. It adds a lot to the build time on mac, and we haven't been bumping the waterfall version in Binaryen recently. By the time we do, perhaps we will have gotten the runtimes build of compiler-rt sorted out.

PR #288, opened 2017-11-20
 Make waterfall use prebuilt Java

aheejin:
Java installed in the buildbots are too old while emscripten uses
closure compiler that requires Java SE 8.0 (version 52) or above.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -193,16 +193,19 @@ NODE_BIN = Executable(os.path.join(WORK_DIR,
 # compiler that requires Java SE 8.0 (version 52) or above
 JAVA_VERSION = '9.0.1'
 
+
 def JavaDir():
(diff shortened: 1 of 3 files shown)


src/build.py
```diff
@@ -586,6 +609,13 @@ def SyncGNUWin32(name, src_dir, git_repo):
   return SyncArchive(GNUWIN32_DIR, name, url)
 
 
+def SyncPrebuiltJava(name, src_dir, git_repo):
+  platform = {'linux2': 'linux', 'darwin': 'osx', 'win32': 'x64'}[sys.platform]
```
dschuff: should win32 be 'windows' instead of 'x64'?

aheejin: 😱 Done

PR #290, opened 2017-11-28
Sleep after writing emscripten configuration file

aheejin:
Each invocation of emcc does sanity check for the current emscripten
cache directory, whose default location is `~/.emscripten_cache`. It
compares the last modification of the configuration file
(`emscripten_config` for asmjs and `emscripten_config_vanilla` for wasm)
with that of the sanity file (`emscripten_config_sanity` for asmjs and
`emscripten_config_vanilla_sanity_wasm` for wasm), and if the
configuration file was modified after the sanity file, deletes the cache
directory and rewrites the sanity file.
https://github.com/kripken/emscripten/blob/a243a941d911698a703fd4e5210a9dbc24bc4e96/tools/shared.py#L512-L533
https://github.com/kripken/emscripten/blob/a243a941d911698a703fd4e5210a9dbc24bc4e96/tools/shared.py#L578-L582

The problem is, emcc thinks the cache is invalid when
`sanity file modification time <= config file modification time`, which
means the cache will be deleted even if two timestamps are equal.
https://github.com/kripken/emscripten/blob/a243a941d911698a703fd4e5210a9dbc24bc4e96/tools/shared.py#L520
This does not happen usually in Linux, but in Mac OS X and Windows, [the
file modification timestamps have only 1 or 2 seconds
resolution](https://docs.python.org/3/library/os.html#os.stat). And
the system library creation by embuilder involves several invocation of
emcc: emcc is invoked per every library (.a or .bc) to build, each of
which also launches multiple emcc processes for each .c file that
belongs to the library. This means it is possible that the cache is
deemed invalid and deleted in the first emcc invocation, but child emcc
processes wlll also delete the cache again, which would then contain
half-built libraries, if less than 1 seconds has elapsed after the the
first deletion.

Modifying the sanity check equation from `<=` to `<` so that the cache
will only be deleted when
`sanity file modification time < config file modification time` does not
seem to be a good solution, because it is also possible that in other
case it is possible that config file creation is followed by sanity file
creation, which warranties the regeneration of sanity file.

This patch just disables checking of sanity file timestamp. The
emscripten cache directory is deleted in the beginning of the function
anyway, so it's ok not to check it for the first emcc call as well.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1117,14 +1117,14 @@ def Emscripten(use_asm):
     # unnecessary anyway, and we are planning to fix it. But this is also
     # causing problems for Mac and Windows build in waterfall. The first emcc
     # process sees the configuration file has changed, deletes cache, and
-    # rewrites the sanity file. Then, Mac OS X and Windows' timestamps for mtime
-    # only have 1~2 seconds resolution, which makes it possible that other emcc
-    # processes launched later by the initial emcc process again think the cache
-    # is invalid and delete the cache again, which would then contain half-built
-    # libraries, if less than 1 seconds has elapsed after the first deletion. So
-    # here we make sure that enough time has elapsed between configuration file
-    # writing and sanity checking, so that it would temporarily work on Mac and
-    # Windows.
+    # rewrites the sanity file. Then, Mac OS X and Windows' timestamps for
+    # mtime only have 1~2 seconds resolution, which makes it possible that
+    # other emcc processes launched later by the initial emcc process again
+    # think the cache is invalid and delete the cache again, which would then
+    # contain half-built libraries, if less than 1 seconds has elapsed after
+    # the first deletion. So here we make sure that enough time has elapsed
+    # between configuration file writing and sanity checking, so that it would
+    # temporarily work on Mac and Windows.
     time.sleep(3)
 
     try:
--- a/src/build.py
+++ b/src/build.py
@@ -1111,6 +1111,7 @@ def Emscripten(use_asm):
       # This depends on binaryen already being built and installed into the
       # archive/install dir.
       os.environ['EM_CONFIG'] = config
+      os.environ['EMCC_SKIP_SANITY_CHECK'] = '1'
       proc.check_call([
           sys.executable, os.path.join(emscripten_dir, 'embuilder.py'),
           'build', 'SYSTEM'])
@@ -1120,6 +1121,7 @@ def Emscripten(use_asm):
       buildbot.Fail()
     finally:
       del os.environ['EM_CONFIG']
+      del os.environ['EMCC_SKIP_SANITY_CHECK']
 
   wrapper = os.path.join(SCRIPT_DIR, 'emcc_wrapper.sh')
   shutil.copy2(wrapper, os.path.join(INSTALL_BIN, 'emcc'))
--- a/src/build.py
+++ b/src/build.py
@@ -1106,12 +1106,32 @@ def Emscripten(use_asm):
     print 'Config file: ', config
     src_config = os.path.join(SCRIPT_DIR, os.path.basename(config))
     WriteEmscriptenConfig(src_config, config)
+
+    # FIXME HACK emcc does sanity check by comparing mtime of the sanity file
+    # with that of emscripten config file. The sanity check process in
+    # emscripten thinks current cache is invalid when 'sanity file's mtime <=
+    # config file's mtime'. embuilder launches emcc for each of system library
+    # we need to build, and each of those emcc process in turn launches several
+    # more emcc processes for each of .c source file included in the library.
+    # Each of these emcc processes runs its own sanity check, which is
+    # unnecessary anyway, and we are planning to fix it. But this is also
+    # causing problems for Mac and Windows build in waterfall. The first emcc
+    # process sees the configuration file has changed, deletes cache, and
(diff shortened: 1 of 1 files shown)


dschuff:
The sanity check has other checks besides those for cache deletion. It would be probably be good to make sure those still run (especially since this builder step is the first time we run emcc, so if something is going to go wrong, it will be here, so this is exactly the place where we want the sanity check).

Does emcc have a mechanism to disable the sanity check in recursive calls of emcc? (or does it just never call emcc recursively? I think it does). If so we could maybe have embuilder explicitly run the sanity check and then disable it when building the libraries. Actually that would probably work even if emcc doesn't disable recursive sanity checks. But we shoudl also maybe make emcc disable recursive sanity checks anyway :)

sbc100:
Seems like there should be fix for this on the emscripten side really.     Perhaps when creating the sanity we could find a way to make sure it is always > than the timestamp of the config file?   (perhaps my setting the mtime explictly, or waiting?)

aheejin:
By the way does anyone know how to embed code snippets within PR? I tried to do [this](https://github.com/blog/2415-introducing-embedded-code-snippets) but they don't appear as embedded snippets. :(

aheejin:
@dschuff emcc does recursively call emcc. At least it does when creating system libraries, the problem in this case. Apparently each system library has its dedicated function in [`system_libs.py`](https://github.com/kripken/emscripten/blob/incoming/tools/system_libs.py), like [`create_compilier_rt`](https://github.com/kripken/emscripten/blob/c56eb77e629d79b9da84e0c193fe7c8a0d203a9e/tools/system_libs.py#L276) for compiler-rt. In case of compiler-rt, [here](https://github.com/kripken/emscripten/blob/c56eb77e629d79b9da84e0c193fe7c8a0d203a9e/tools/system_libs.py#L287) is where it launches several other emcc processes to compile each .c file that belongs to compiler-rt. Other library-building functions have the similar structure.

I'm not entirely sure if this is the only place where launches other emcc processes. (Probably not.) But anyway, for this system library building part, to disable this `EMCC_SKIP_SANITY_CHECK` option when emcc launches other emcc processes, the only place I can think of is this [`calculate`](https://github.com/kripken/emscripten/blob/c56eb77e629d79b9da84e0c193fe7c8a0d203a9e/tools/system_libs.py#L36) function, which calls all other library-building functions ([`create_compiler_rt`](https://github.com/kripken/emscripten/blob/c56eb77e629d79b9da84e0c193fe7c8a0d203a9e/tools/system_libs.py#L276), [`create_libcxx`](https://github.com/kripken/emscripten/blob/c56eb77e629d79b9da84e0c193fe7c8a0d203a9e/tools/system_libs.py#L197), ... etc.) The full list seems to be [here](https://github.com/kripken/emscripten/blob/c56eb77e629d79b9da84e0c193fe7c8a0d203a9e/tools/system_libs.py#L436-L442) But as you can see, the [`calculate`](https://github.com/kripken/emscripten/blob/c56eb77e629d79b9da84e0c193fe7c8a0d203a9e/tools/system_libs.py#L36) function is long and possibly has many exit points. One option can be to wrap the whole function with big try-finally and make sure we unset the `EMCC_SKIP_SANITY_CHECK` variable at the end?

And I like the idea of doing the sanity check at the beginning of embuilder and disable the `EMCC_SKIP_SANITY_CHECK` option for the rest of the embuilder process.

aheejin:
@sbc100 For waiting (for a 2~5 seconds), I thought about doing that in the waterfall emscripten build actually. That would work ok for the waterfall at least. But inserting the waiting logic into emscripten makes every emscripten process few seconds longer, which might not be acceptable.

I'm not sure if we can set the mtime explicitly. And even if we can, wouldn't that interfere with the 'normal' mtime setting logic that's used elsewhere?

dschuff:
I think the right solution is probably to create some shared function `run_emcc` and make it call emcc with some flag or modified environment, and replace all the manual invocations of emcc (at least those in emscripten proper, if not the tests too). 

There are way to many flavors and variations of `Popen([shared.Python, 'some_random_thing'])` in emscripten. But that's a rant for another PR.

dschuff:
Doing that will also make everything faster, since it will avoid redundant sanity checks.

aheejin:
Yeah I agree, I was kinda hoping for not having to do that though... 😞 Anyway, then can we land this quick fix to make waterfall green again before that? If the problem is this disables the whole checks controlled by the flag, we can sleep for 3 seconds after writing a config file and before proceeding to embuilder, which gives enough time for timestamp difference in Mac & Windows. Looks like a horrible hack though.

sbc100:
If you want to land a quick fix wouldn't it be less risky to simple do "sleep(1)" right after we create the config files?

Skipping the sanity checks seems like a comparatively big change.

dschuff:
I'm ok with a temporary hack in the waterfall if we also fix it the right way. But I actually do think it's better to just sleep for 3 seconds; firstly, either way is a horrible hack, and sleeping *looks* more like the horrible hack that it is, and also for the reason that I mentioned before; this is the most likely place for an actual sanity check failure.

sbc100:
Sorry to be nit picky but why not 1 second?

aheejin:
To be safe. The [note](https://docs.python.org/3/library/os.html#os.stat_result.st_ctime_ns) in this page says on Windows systems using the FAT or FAT32 file systems, `st_mtime` has 2-second resolution.

PR #292, opened 2017-11-30
Install arch/generic/bits/*.h from musl as well as arch/wasm32/bits/*.h

sbc100:
These should alwasy have been installed but it recently started
mattering becasue I removed bits/mman.h.

This mimics more closely what the musl Makefile does on `install`

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1221,8 +1221,10 @@ def Musl():
 
     CopyTree(os.path.join(MUSL_SRC_DIR, 'include'),
              os.path.join(INSTALL_SYSROOT, 'include'))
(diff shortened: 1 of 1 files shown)


aheejin:
Aren't there overlapping files? In case there are, overwriting generic ones with wasm32 ones is your intention?

sbc100:
Yes,  we want the wasm32 ones to take precedence, and the auto-generated ones to be even higher precedence.

sbc100:
The musl-side change is now landed: https://github.com/jfbastien/musl/pull/31

aheejin:
Aren't there overlapping files? In case there are, overwriting generic ones with wasm32 ones is your intention?

PR #297, opened 2017-12-06
Update CMake toolchain file for Windows and CMAKE_AR

dschuff:
1. Use exe suffix to find clang on Windows
2. Only set the tools if they are not already overridden (for some reason this is required for CMAKE_AR and CMAKE_RANLIB to work).
3. Pass --target by default for proper auto-detection of the target platform.


dschuff:
This, combined with a CMake update, will fix windows compiler-rt.

src/wasm_standalone.cmake
```diff
@@ -9,12 +9,28 @@ set(CMAKE_SYSTEM_PROCESSOR wasm32)
...
+        set(EXE_SUFFIX "")
+endif()
+if ("${CMAKE_C_COMPILER}" STREQUAL "")
+  set(CMAKE_C_COMPILER ${WASM_SDKROOT}/bin/clang${EXE_SUFFIX})
+endif()
+if ("${CMAKE_CXX_COMPILER}" STREQUAL "")
+  set(CMAKE_CXX_COMPILER ${WASM_SDKROOT}/bin/clang++${EXE_SUFFIX})
+endif()
+if ("${CMAKE_AR}" STREQUAL "")
+  set(CMAKE_AR ${WASM_SDKROOT}/bin/llvm-ar${EXE_SUFFIX} CACHE FILEPATH "llvm ar")
```
sbc100: What are the last 3 args here doing and why are they only needed for AR and RANLIB?

src/wasm_standalone.cmake
```diff
@@ -9,12 +9,28 @@ set(CMAKE_SYSTEM_PROCESSOR wasm32)
...
+endif()
+if ("${CMAKE_RANLIB}" STREQUAL "")
+ set(CMAKE_RANLIB ${WASM_SDKROOT}/bin/llvm-ranlib${EXE_SUFFIX} CACHE FILEPATH "llvm ranlib")
+endif()
 set(CMAKE_C_COMPILER_TARGET ${triple})
 set(CMAKE_CXX_COMPILER_TARGET ${triple})
+set(CMAKE_C_FLAGS --target=${triple})
+set(CMAKE_CXX_FLAGS --target=${triple})
```
sbc100: Why wasn't this needed before?

src/wasm_standalone.cmake
dschuff: I'm not 100% sure why it's different; I cargo-culted this from emscripten (https://github.com/kripken/emscripten/blob/incoming/cmake/Modules/Platform/Emscripten.cmake#L82) The CACHE bit causes the path to go into CMakeCache.txt at configure time (and there's some interaction between that and when the toolchain cmake file is evaluated, which happens more than once). But I don't know why ar is special compared to the compiler.

src/wasm_standalone.cmake
dschuff: It affects CMake's detection of the compiler's default target. If you don't put that, CMake thinks the compiler is targeting Windows.

dschuff: (anyway, I think it was needed before; we just didn't have it. We don't want to require the user to pass --target themselves). What I think we should actually do is make a symlink to name the compiler `wasm32-clang` as other gnu-style cross-compilers do, and then it will automatically get the target right. We'll just have to coordinate that with an emscripten change.

PR #298, opened 2017-12-06
Update CMake to 3.7.2

dschuff:
This version contains a fix to the Ninja generator that properly recognizes gnu-style compilers on Windows and creates response files with the correct file path escaping.


dschuff:
Oh, this depends on #297, will land that first.

PR #299, opened 2017-12-08
Compile JSC on Windows and Linux as well

mathiasbynens:
CC @dschuff @jfbastien

Ref. https://github.com/v8/web-tooling-benchmark/pull/29#discussion_r155831751

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -948,7 +948,7 @@ def Jsc():
     command = ['xcrun', PREBUILT_CMAKE_BIN, '-Wno-dev',
                '..', '-G', 'Ninja',
                '-DCMAKE_BUILD_TYPE="Release"',
-               '-DPORT=Mac',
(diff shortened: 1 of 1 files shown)


sbc100:
Won't you need to remove the `no_windows` and `no_linux` args from line 1430? 

Also, you'll need to make the use of xcrun conditional won't you?  (since it doesn't exist on non-mac).  Maybe xcrun isn't needed?  Its not used for other cmake runs in this file.

dschuff:
I'll give this a try locally.

mathiasbynens:
> Won't you need to remove the `no_windows` and `no_linux` args from line 1430?

Yes! Done.

I don’t feel qualified to deal with the `xcrun` issue, to be honest. Feel free to amend this patch.

dschuff:

PR #300, opened 2017-12-08
Remove forcing of MSVS 2015 toolchain

dschuff:
This dates back to when 2013 was the other option, but now only 2017 is
supported.


jfbastien:
2017 is only valid for a few more days though 🙃

dschuff:
Fortunately this just removes the pinning and lets the chrome tooling run whatever it wants.

dschuff:
so when can I get C++18? Is there a rule about versions being monotonic?

jfbastien:
> so when can I get C++18?

Hey 17 only just because a full ISO standard this week! Wait at least 3 years for C++20.

> Is there a rule about versions being monotonic?

Hmmm... I don't think there is.

PR #302, opened 2017-12-14
Run emscripten tests in node.js instead of d8

dschuff:
Proof of concept to run bare and emscripten tests in node.js.

Everything seems to work but Interestingly there seems to be unexpected successes (compared to d8) in the emscripten torture tests (both asm2wasm and emwasm)

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1514,7 +1514,7 @@ def TestBare():
   for opt in BARE_TEST_OPT_FLAGS:
     ExecuteLLVMTorture(
         name='d8-lld',
-        runner=NODE_BIN,
+        runner=Executable(os.path.join(INSTALL_BIN, 'd8')),
         indir=GetTortureDir('lld', opt),
         fails=RUN_KNOWN_TORTURE_FAILURES,
         attributes=common_attrs + ['d8', 'lld'],
@@ -1546,7 +1546,7 @@ def TestBare():
   for opt in BARE_TEST_OPT_FLAGS:
     ExecuteLLVMTorture(
         name='d8',
-        runner=NODE_BIN,
+        runner=Executable(os.path.join(INSTALL_BIN, 'd8')),
         indir=GetTortureDir('wat2wasm', opt),
         fails=RUN_KNOWN_TORTURE_FAILURES,
         attributes=common_attrs + ['d8'],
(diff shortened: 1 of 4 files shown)


dschuff:
@jgravelle-google we could use this as a temporary workaround if a real fix for the current d8 issue is not imminent. We'd probably want to add an attribute to the test exclusions and update them accordingly. To get extra test coverage we could switch some of our d8 tests over to node, or add some extra runs of the torture tests.

jgravelle-google:
We should probably do this, it's a pretty uncontroversial way to green the waterfall back up.

Some unexpected successes seems odd.
Is everything unexpectedly succeeding? When I threw together something quickly locally to run node, it ran all the tests without actually returning the exit code to the shell, so the test would trap, and node would return 0, making it an unexpected success.
But if it's only a few, then idk.

src/build.py
```diff
@@ -1514,7 +1514,7 @@ def TestBare():
   for opt in BARE_TEST_OPT_FLAGS:
     ExecuteLLVMTorture(
         name='d8-lld',
-        runner=Executable(os.path.join(INSTALL_BIN, 'd8')),
+        runner=NODE_BIN,
```
jgravelle-google: Strictly speaking we don't need to run the bare tests in node, just asm and emwasm. Although it's maybe better to run them all in the same shell? I have no preference here.

dschuff:
It's not everything that's unexpectedly succeeding. I haven't confirmed but I think it might be related to this text I'm seeing in the logs:
`(node:248561) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.`
Previous to that is a stacktrace from abort(), but it looks a little weird, so I'm looking at that.

sbc100:
Isn't it a major downside here that we are not building node from source and that node lags d8.   Basically it means this waterfall will not longer be testing against v8 changes as the happen right?  Seems like that goes against the original purpose of this waterfall which was to be an integration test for all the things.

dschuff:
Short answer: yes, and that's why this would be a temporary workaround, until we figure out what to do about the async compile thing (see https://github.com/kripken/emscripten/pull/5937 and https://bugs.chromium.org/p/v8/issues/detail?id=7212 and https://github.com/WebAssembly/threads/issues/52)

dschuff:
With https://github.com/kripken/emscripten/pull/5948 and this PR, the bot should be green again (at least, it seem working locally for me).

PR #308, opened 2018-01-10
Add ability to completely exclude torture tests from test pipeline

dschuff:
The test in question has UB and is optimized into an infinite loop. On mac and
linux the python rlimit mechanism is used to kill the program after timeout, but
there's no equivalent functionality for Windows currently. So we just exclude
the test entirely (this also makes everything run faster on mac and linux since
we aren't waiting uselessly for the timeout).


dschuff:
For some reason, flake8's restriction to 79 chars instead of 80 still makes me stabby.

jgravelle-google:
https://github.com/petersn/six-char-max clearly

sbc100:
Thank you!

PR #310, opened 2018-01-10
Also use exclusions with emcc-built torture tests

dschuff:


dschuff:
Gonna "TBR" this as it fixes the waterfall (should have been in the last commit)

PR #311, opened 2018-01-11
Add libcxx and libcxxabi builds

sbc100:


src/build.py
```diff
@@ -635,6 +639,10 @@ def NoSync(*args):
            LLVM_MIRROR_BASE + 'clang.git'),
     Source('compiler-rt', COMPILER_RT_SRC_DIR,
            LLVM_MIRROR_BASE + 'compiler-rt.git'),
+    Source('libcxx', LIBCXX_SRC_DIR,
+           LLVM_MIRROR_BASE + 'libcxx.git'),
+    Source('libcxxabi', LIBCXXABI_SRC_DIR,
```
dschuff: This is fine for now but at some point we'll want our own fork of libcxxabi so that @aheejin can check in the experimental EH code before it goes upstream; sort of like we did with lld.

src/build.py
```diff
@@ -1190,6 +1198,51 @@ def CompilerRT():
...
+def LibCXX():
+  buildbot.Step('libcxx')
+  if os.path.isdir(LIBCXX_OUT_DIR):
+    Remove(LIBCXX_OUT_DIR)
+  Mkdir(LIBCXX_OUT_DIR)
+  cc_env = BuildEnv(LIBCXX_SRC_DIR, bin_subdir=True)
+  command = [PREBUILT_CMAKE_BIN, '-G', 'Ninja', os.path.join(LIBCXX_SRC_DIR),
+             '-DCMAKE_CXX_COMPILER_WORKS=ON',
+             '-DCMAKE_C_COMPILER_WORKS=ON',
+             '-DLIBCXX_HAS_PTHREAD_API=ON',
```
dschuff: Is this going to result in a libcxx that requires pthreads on by default? That's probably not what we want, at least for now.

sbc100: You're right.. I found ENABLE_THREADS=OFF.

src/build.py
jfbastien: LMK when you need a repo.

sbc100:
Should be good to go now.  PTAL.

dschuff:
The libcxx build doesn't quite work: https://logs.chromium.org/v/?s=chromium%2Fbb%2Fclient.wasm.llvm%2Flinux%2F28011%2F%2B%2Frecipes%2Fsteps%2Flibcxxabi%2F0%2Fstdout

It's because CMake's version check uses the atomic header to determine if the c++ stdlib is new enough:
```
Performing C++ SOURCE FILE Test LLVM_NO_OLD_LIBSTDCXX failed with the following output:
Change Dir: /usr/local/google/home/dschuff/s/wasm-waterfall/src/work/libcxxabi-out/CMakeFiles/CMakeTmp

Run Build Command:"/usr/local/google/home/dschuff/depot_tools/ninja" "cmTC_54b43"
[1/2] Building CXX object CMakeFiles/cmTC_54b43.dir/src.cxx.obj
FAILED: CMakeFiles/cmTC_54b43.dir/src.cxx.obj 
/usr/local/google/home/dschuff/s/wasm-waterfall/src/work/wasm-install/bin/clang++  --target=wasm32-unknown-unknown-wasm --sysroot=/usr/local/google/home/dschuff/s/wasm-waterfall/src/work/wasm-install/sysroot    -DLLVM_NO_OLD_LIBSTDCXX  -std=c++0x -o CMakeFiles/cmTC_54b43.dir/src.cxx.obj -c /usr/local/google/home/dschuff/s/wasm-waterfall/src/work/libcxxabi-out/CMakeFiles/CMakeTmp/src.cxx
In file included from /usr/local/google/home/dschuff/s/wasm-waterfall/src/work/libcxxabi-out/CMakeFiles/CMakeTmp/src.cxx:2:
/usr/local/google/home/dschuff/s/wasm-waterfall/src/work/wasm-install/sysroot/include/c++/v1/atomic:553:2: error: <atomic> is not supported on this single threaded system
#error <atomic> is not supported on this single threaded system
 ^
1 error generated.
ninja: build stopped: subcommand failed.

Source file was:

#include <atomic>
std::atomic<float> x(0.0f);
int main() { return (float)x; }
```
That seems... dubious :-/
I guess for now we can override that check. Going forward we'll probably need to figure out a toolchain story for the threads vs no-threads feature.

sbc100:
Disabled libcxxabi for now

sbc100:
I wonder why this doesn't fail for me locally.

PR #314, opened 2018-01-17
Include lld and compiler-rt when syncing to llvm rev

sbc100:


src/build.py
```diff
@@ -750,8 +750,8 @@ def get_rev(rev_name):
   primary = LLVM_SRC_DIR if SCHEDULER == 'llvm' else CLANG_SRC_DIR
```
dschuff: This function should probably be renamed now.

PR #316, opened 2018-01-18
Add support for some of the C++ gcc testsuite

sbc100:
The gcc test suite is spread over several directories.  There are also duplicate filenames across
those directories.  This means we can't just use the test basename for complication output.
I looked into re-creating the nested structure in the output tree, but in the end it was simpler
to convert the full path into single filename (essentially replacing `\` with `__`). Hopefully this
is not too ugly!

### DIFF
--- a/src/test/lld_known_gcc_test_failures.txt
+++ b/src/test/lld_known_gcc_test_failures.txt
@@ -8,3 +8,12 @@ va-arg-pack-1.c.o
 # undefined symbol: link_error
 # Don't care. The test case is faulty. link_error() does not exist.
 medce-1.c.o O0
+
+asan__interception-test-1.C.o  # Undefined symbol: __interceptor_strtol
+tree-ssa__pr20458.C.o  # Undefined symbol: std::locale::locale
+
+# Crash in lld
+warn__weak1.C.o
+
+# Untriaged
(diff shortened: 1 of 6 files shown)


src/test/lld_known_gcc_test_failures.txt
```diff
@@ -8,3 +8,12 @@ va-arg-pack-1.c.o
 # undefined symbol: link_error
 # Don't care. The test case is faulty. link_error() does not exist.
 medce-1.c.o O0
+
+asan__interception-test-1.C.o  # Undefined symbol: __interceptor_strtol
```
jfbastien: Oooh that would be neat to support!

src/testing.py
```diff
@@ -23,6 +23,7 @@
 
 import proc
 
+SINGLE_THREADED = False
```
jfbastien: Why?

dschuff: I'm guessing it becomes easier to debug, or to see the output?

jfbastien: Yeah but it'll run the tests way slower!

dschuff: Sure, that's why it's false by default.

sbc100: I'll split this out into separate change.

PR #318, opened 2018-01-19
Add a simple way runs tests single threaded during development

sbc100:
When working on waterfall code the thread pool can sometimes
make this very hard to debug, hiding output and errors in some
cases.

### DIFF
--- a/src/testing.py
+++ b/src/testing.py
@@ -26,6 +26,7 @@ import proc
 # Set to True to disable execution via thread pool
 single_threaded = False
 
+
 class Result:
   """Result from a single test that was run."""
 
--- a/src/testing.py
+++ b/src/testing.py
@@ -23,7 +23,8 @@ import sys
 
 import proc
 
-SINGLE_THREADED = False
+# Set to True to disable execution via thread pool
+single_threaded = False
 
 class Result:
   """Result from a single test that was run."""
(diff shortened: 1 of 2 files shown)


src/testing.py
```diff
@@ -229,11 +230,14 @@ def execute(tester, inputs, fails, exclusions=None, attributes=None):
     input_expected_failures = parse_exclude_files(fails, attributes)
   else:
     input_expected_failures = []
-  pool = multiprocessing.Pool()
   sys.stdout.write('Executing tests.')
-  results = sorted(pool.map(tester, inputs))
-  pool.close()
-  pool.join()
+  if SINGLE_THREADED:
+    results = map(tester, inputs)
```
jfbastien: `sorted` here as well?

src/testing.py
```diff
@@ -23,6 +23,7 @@
 
 import proc
 
+SINGLE_THREADED = False
```
jfbastien: Can you pipe this in from the command line instead?

src/testing.py
sbc100: Done

src/testing.py
sbc100: Done... although still using a global.. otherwise we need to patch all the callsites..

PR #320, opened 2018-01-20
Support multiple modes for Emscripten testsuites

aheejin:


src/build.py
```diff
@@ -257,6 +257,7 @@ def JavaBinDir():
 # Optimization levels
 BARE_TEST_OPT_FLAGS = ['O0', 'O2']
 EMSCRIPTEN_TEST_OPT_FLAGS = ['O0', 'O3']
+EMSCRIPTEN_TESTSUITE_MODES = ['binaryen0', 'binaryen2']
```
sbc100: Out of interest what are the two different modes?

aheejin: https://github.com/kripken/emscripten/blob/941e9de0946077d835a40556048af1de4e4e2abe/tests/test_core.py#L7517-L7523

dschuff:
The code LGTM. Do we need to make binaryen0 xfail for now?

aheejin:
Not sure. Last time I checked there were two new failures on this mode; should we look at it and fix it now or make them xfail?

dschuff:
We don't need to block this on fixing everything. But I'd rather not check something into the waterfall that will make it red?

aheejin:
Rather than making the whole mode xfail, how about landing this as-is and adding the failing two tests to the expected failure list?

aheejin:
Oh no, nevermind, there's no failure expectation list for emscripten test suite.. I was confused

aheejin:
Hmm, currently there's no way to make only a specific mode or compilation option xfail. We should either make the whole emscripten testsuite warn_only or not. I can add the option of making only one of them xfail, but not sure it's worth it, because I think we are gonna fix them anyway. Hmm.

aheejin:
For the record, the failing tests with binaryen0 are these two:
- test_bswap64
- test_utf32
We can land this after we fix these then.

sbc100:
Any interest in still landing this? 

I landed a kind of support for muitple modes but maybe your prefer this method?  See #393

aheejin:
Oh the reason I was unable to land this was because there were failing tests. At the time of this PR it was `test_bswap64` and `test_utf32`, but now I just ran `binaryen0` again and `test_float_builtins` and `test_lower_intrinsics` are failing. Maybe we should fix them, and add `binaryen0` to what you did in #393?

By the way, currently what's the main diff between `binaryen` and `wasmobj` tests? I think both of them use lld, right? Does `binaryen` still use `llvm-link`?

sbc100:
> Oh the reason I was unable to land this was because there were failing tests. At the time of this PR it was `test_bswap64` and `test_utf32`, but now I just ran `binaryen0` again and `test_float_builtins` and `test_lower_intrinsics` are failing. Maybe we should fix them, and add `binaryen0` to what you did in #393?
> 
> By the way, currently what's the main diff between `binaryen` and `wasmobj` tests? I think both of them use lld, right? Does `binaryen` still use `llvm-link`?

The different is the object files.  One uses bitcode as object format and the other uses wasm object files.

sbc100:
Still want to land this or should we close?

aheejin:
It's so out of date anyway, so let's close.

PR #325, opened 2018-01-27
Add jsvu step

mathiasbynens:
This patch adds a step that installs [jsvu](https://github.com/GoogleChromeLabs/jsvu). For now it only does so on Windows (where jsvu then installs a `chakra` binary) and Mac (where jsvu then installs a `javascriptcore` binary), but this can easily be extended in the future.

Hopefully this can replace the complex build systems for these engines and reduce the maintenance cost of this project.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1548,7 +1548,7 @@ def AllBuilds(use_asm=False):
       Build('llvm', LLVM),
       Build('v8', V8),
       Build('jsc', Jsc, no_windows=True, no_linux=True),
-      Build('jsvu', Jsvu, no_linux=True),
+      Build('jsvu', Jsvu, no_windows=True, no_linux=True),
       Build('wabt', Wabt),
       Build('ocaml', OCaml, no_windows=True),
       Build('spec', Spec, no_windows=True),
--- a/src/build.py
+++ b/src/build.py
@@ -59,6 +59,8 @@ FASTCOMP_SRC_DIR = os.path.join(WORK_DIR, 'emscripten-fastcomp')
 GCC_SRC_DIR = os.path.join(WORK_DIR, 'gcc')
 GCC_TEST_DIR = os.path.join(GCC_SRC_DIR, 'gcc', 'testsuite')
 
+JSVU_DIR = os.path.join(WORK_DIR, 'jsvu')
+
 V8_SRC_DIR = os.path.join(WORK_DIR, 'v8', 'v8')
 JSC_SRC_DIR = os.path.join(WORK_DIR, 'jsc')
 WABT_SRC_DIR = os.path.join(WORK_DIR, 'wabt')
@@ -199,10 +201,16 @@ PREBUILT_CMAKE_BASE_NAME = 'cmake-%s-%s-%s' % (PREBUILT_CMAKE_VERSION,
 PREBUILT_CMAKE_DIR = os.path.join(WORK_DIR, PREBUILT_CMAKE_BASE_NAME)
 PREBUILT_CMAKE_BIN = os.path.join(PREBUILT_CMAKE_DIR, CMakeBinDir(), 'cmake')
 
-NODE_BIN = Executable(os.path.join(WORK_DIR,
-                                   NODE_BASE_NAME + NodePlatformName(),
-                                   'bin', 'node'))
+NODE_DIR = os.path.join(WORK_DIR, NODE_BASE_NAME + NodePlatformName())
+NODE_BIN_DIR = os.path.join(NODE_DIR, 'bin')
+# `npm` uses whatever `node` is in `PATH`. To make sure it uses the
+# Node.js version we want, we prepend `NODE_BIN_DIR` to `PATH`.
+os.environ['PATH'] = NODE_BIN_DIR + os.pathsep + os.environ['PATH']
+NODE_BIN = Executable(os.path.join(NODE_BIN_DIR, 'node'))
+NPM_BIN = Executable(os.path.join(NODE_BIN_DIR, 'npm'))
 
+JSVU_BIN = Executable(os.path.join(JSVU_DIR,
+                                   'node_modules', 'jsvu', 'cli.js'))
 
 # Java installed in the buildbots are too old while emscripten uses closure
 # compiler that requires Java SE 8.0 (version 52) or above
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -965,6 +965,36 @@ def V8():
...
+      os_id = 'windows64'
+      js_engines = 'chakra'
+    elif IsMac():
+      os_id = 'mac64'
+      js_engines = 'javascriptcore'
+    else:
+      return
+
+    # https://github.com/GoogleChromeLabs/jsvu#installation
+    proc.check_call(['npm', 'install', 'jsvu', '-g'])
```
dschuff: I'm guessing this isn't going to work as-is because npm won't be in the PATH. Take a look at how we do `NODE_BIN` up around line 200.

src/build.py
```diff
@@ -965,6 +965,36 @@ def V8():
...
+
+    # https://github.com/GoogleChromeLabs/jsvu#integration-with-non-interactive-environments
+    proc.check_call(['jsvu',
```
dschuff: likewise here. use an explicit path to JSVU.

dschuff: Actually is `jsvu` a binary or a shell script, or a JS file that we have to run with node, or what?

src/build.py
mathiasbynens: Done.

src/build.py
mathiasbynens: Done. Note that `npm install -g` puts `jsvu` [wherever Node is installed](https://docs.npmjs.com/files/folders#tldr).

src/build.py
```diff
@@ -202,6 +202,12 @@ def CMakeBinDir():
 NODE_BIN = Executable(os.path.join(WORK_DIR,
                                    NODE_BASE_NAME + NodePlatformName(),
                                    'bin', 'node'))
+NPM_BIN = Executable(os.path.join(WORK_DIR,
+                                  NODE_BASE_NAME + NodePlatformName(),
```
dschuff: maybe we should just refactor the directory part into `NODE_DIR` and then have `NODE_BIN = Executable(os.path.join(NODE_DIR, 'node'))`. Would probably even be fewer lines.

mathiasbynens: Ack

dschuff:
I tested this on my laptop and got the following error:
```
@@@BUILD_STEP jsvu@@@
subprocess.check_call(`/Users/dschuff/code/waterfall/src/work/node-v8.9.3-darwin-x64/bin/npm install jsvu`, cwd=`/Users/dschuff/code/waterfall/src/work/jsvu`)
ERROR: npm is known not to run on Node.js v4.4.4
Node.js 4 is supported but the specific version you're running has
a bug known to break npm. Please update to at least 4.7.0 to use this
version of npm. You can find the latest release of Node.js at https://nodejs.org/
```
I apparently have node 4.4.4 installed in /usr/local/bin so it looks like npm (even though it looks like npm is the correct one) is finding the wrong node. Do you know if there's a way to make npm do the right thing (flag, config file, etc)? Or do we need to inject the node path into PATH?

mathiasbynens:
I suppose you’re getting the same old Node when running `node -v`?

Does the build environment have Node.js v4 installed as well? (Before submitting the PR I verified that https://github.com/WebAssembly/waterfall/blob/88fe2f1b2d3d3b18d1c7104958bb2c26b2f4b58c/src/build.py#L191-L193 installs a recent-enough Node.js version.)

(Note that @zkat’s [pro tip](https://twitter.com/maybekatz/status/958157474397171712) to run `npm install node@8` before installing `jsvu` doesn’t help here, as it would still trigger the warning when that initial command is invoked.)

AFAICT npm-cli just uses whatever `node` binary is in the `PATH`:

* https://github.com/npm/npm/blob/0cc9d89ed2d46745f91d746fda9d205fd39d3daa/bin/npm#L12
* https://github.com/npm/npm/blob/0cc9d89ed2d46745f91d746fda9d205fd39d3daa/bin/npm-cli.js#L1

Adding the `node` path to `PATH` would solve the problem.

dschuff:
I don't think the bots have node installed at all, but I want the script to be hermetic and not depend on that. So we can just put yet another thing in the path.

mathiasbynens:
Patch updated.

dschuff:
So this still doesn't work on Windows because as I mentioned in chat earlier, the install of Node that we use on Windows doesn't actually have npm (the zipfile version of the node/npm package isn't installable because it has a super deep directory hierarchy and runs into path length limitations). So this would still be usable on mac, but until someone fixes that we should just disable this step for Windows.

src/build.py
```diff
@@ -1508,6 +1548,7 @@ def AllBuilds(use_asm=False):
       Build('llvm', LLVM),
       Build('v8', V8),
       Build('jsc', Jsc, no_windows=True, no_linux=True),
+      Build('jsvu', Jsvu, no_linux=True),
```
dschuff: we can just add no_windows=True here for now too.

jfbastien: A Window with no vue? Or a Vue with no window?

mathiasbynens: Ah, so you *don’t* want to install Chakra on Windows yet? Sure, I can add `no_windows=True` to this patch, and we can then remove it in a follow-up commit to add Chakra support.

mathiasbynens: Done. I’ve left the other Windows-specific lines (`if isWindows() …`) intact so we can easily enable support later.

dschuff: I do *want* to install Chakra, it just won't work until someone figures out how to get npm on the bot :) Leaving the windows-specific logic in the function is fine.

mathiasbynens: I’m confused — the bots install a Node.js which comes with npm. See how I set `NPM_BIN` as part of this patch. So, doesn’t this already work?
 https://github.com/WebAssembly/waterfall/pull/325/files#diff-54dfd4a47bc5343af53bfc5a616038f0R209

dschuff: On Windows, the node.js is just bare node.exe without npm (see https://github.com/WebAssembly/waterfall/blob/master/src/build.py#L596). If you download the official zipfile package from Nodejs.org (https://nodejs.org/dist/v8.9.4/node-v8.9.4-win-x64.zip) and attempt to extract it in your waterfall working directory, the extraction will fail because of the ridiculously deep directory structure. Until someone figures out a way to work around that, npm won't work on windows.

mathiasbynens:
Can I track the Windows LLVM build fixing somewhere? Is there an open issue for it?

dschuff:
Hey, it looks fixed now, must have been an upstream issue.

PR #329, opened 2018-02-06
[WIP] Switch to using CMake to build musl

sbc100:


src/build.py
```diff
@@ -1267,47 +1266,31 @@ def LibCXXABI():
...
-    # Note the failure but allow the build to continue.
-    buildbot.Fail()
+  command = [PREBUILT_CMAKE_BIN, '-G', 'Ninja', MUSL_SRC_DIR,
+             '-DCMAKE_TOOLCHAIN_FILE=' + CMAKE_TOOLCHAIN_FILE]
+
+  proc.check_call(command, cwd=MUSL_OUT_DIR, env=cc_env)
+  proc.check_call(['ninja', '-v'], cwd=MUSL_OUT_DIR, env=cc_env)
+  proc.check_call(['ninja', 'install'], cwd=MUSL_OUT_DIR, env=cc_env)
+
+  #CopyLibraryToSysroot(os.path.join(MUSL_OUT_DIR, 'crt1.o'))
```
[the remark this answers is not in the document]
sbc100: No, `ninja install` should install it.  I meant to remove this line.

src/build.py
```diff
@@ -1267,47 +1266,31 @@ def LibCXXABI():
...
-    Mkdir(cmake_path)
-    shutil.copy2(os.path.join(SCRIPT_DIR, 'wasm_standalone.cmake'),
-                 CMAKE_TOOLCHAIN_FILE)
+  cc_env = BuildEnv(MUSL_OUT_DIR, use_gnuwin32=True)
 
-  except proc.CalledProcessError:
-    # Note the failure but allow the build to continue.
-    buildbot.Fail()
+  command = [PREBUILT_CMAKE_BIN, '-G', 'Ninja', MUSL_SRC_DIR,
+             '-DCMAKE_TOOLCHAIN_FILE=' + CMAKE_TOOLCHAIN_FILE]
```
dschuff: I don't think there's any way we can get around using `-DCMAKE_C_COMPILER_WORKS=ON` here, otherwise the tests will always fail to link and you'll error. at least that's what I'm seeing.

sbc100: I wonder why this is working for me ?  I'm pretty sure I did a clean build..

dschuff: So I just tried it again; If I build with no musl-out and a clean wasm-install dir, I get errors. If you just set `CMAKE_C_COMPILER_WORKS=ON`, then the build technically succeeds but all the compiler feature checks fail (due to the same link failure) which isn't ideal. I found that if I set  '-`DCMAKE_EXE_LINKER_FLAGS=-nostdlib -Wl,--allow-undefined'` then all the feature checks link successfully and seem to have the right output. Unfortunately that doesn't work when you add it to the CMakeLists.txt file, it has to be on the command line for some reason.

dschuff: (btw the --allow-undefined thing is needed because if you use `-nostdlib- to avoid linking `-lc` then you have `_start` undefined. That's kind of an ugly hack but I haven't found a better way to make the cmake checks work as intended. Also if you add the linker flags then you can remove `CMAKE_C_COMPILER_WORKS` from the musl cmake file.

sbc100: I ended up setting CMAKE_C_COMPILER_WORKS=OFF in musl's CMakeLists.txt .. which I think makes some sense.

dschuff: (I assume you meant ON...) Right, but all the feature tests still fail if you only do that.

sbc100: I mean `=ON`.  ￼See https://github.com/jfbastien/musl/pull/45

PR #333, opened 2018-02-21
Update expectations

sbc100:


jgravelle-google:
For future historians, C++ spec (N4296), § 5.3.4.10:

> An implementation is allowed to omit a call to a replaceable global allocation function (18.6.1.1, 18.6.1.2). When it does so, the storage is instead provided by the implementation or provided by extending the allocation of another new-expression

PR #335, opened 2018-03-03
Run torture tests through emwasm with lld

sbc100:


src/compile_torture_tests.py
```diff
@@ -77,6 +77,8 @@ def run(cc, cxx, testsuite, sysroot_dir, fails, exclusions, out, config, opt):
                  '--sysroot=%s' % sysroot_dir],
       # Binaryen's native-wasm method uses the JS engine's native support for
       # wasm rather than interpreting the wasm with wasm.js.
+      'binaryen-lld': ['-s', 'WASM=1', '-s',
+                       'BINARYEN_METHOD="native-wasm"'],
```
dschuff: native-wasm is the default method and we don't test anything else anymore, so we can probably just leave this flag out.

PR #338, opened 2018-03-10
Update lld known failures

dschuff:
warn__weak1.C.o works now


dschuff:
TBR, /cc @sbc100 :+1:

sbc100:
Thanks to @NWilson! 

I think it was https://reviews.llvm.org/D44028 that fixed it.

PR #339, opened 2018-03-10
Update expectations

sbc100:
I think this is due to recert wasm-emscripten-finalize changes


sbc100:
I think what is happening is that finalize itself now only generates warnings for type mismatches.. but in O3 these turn into parse errors.

PR #341, opened 2018-03-13
Update toolchain files and CMake builds for more correct feature detection

dschuff:
CMake feature detection for library builds is currently failing due to linking issues with incomplete toolchains, and in some cases may not be using the correct triple (since clang currently defaults to x86 rather than wasm).
This PR adds a symlink called `wasm32-clang`, a standard way to make clang default to a particular triple. This (combined with adding `-nostdlib++` seems to make feature detection work properly for compiler-rt and libcxx.

It also gives the CMake configuration a less-generic name (WACK: the WebAssembly Clang Kit) and makes it more clear that the toolchain built by the waterfall is just one particular experimental configuration but not the only possible one. It's also easier to refer to a particular name instead always saying "the toolchain that the waterfall builds".


dschuff:
@jgravelle-google want to see if this makes things work for you?

jgravelle-google:
>@jgravelle-google want to see if this makes things work for you?

To clarify: my issue is that it *does* work for me locally, so I can't evaluate any potential fixes :/

src/cmake/Modules/Platform/Wack.cmake
```diff
@@ -0,0 +1,29 @@
+# WACK (WebAssembly Clang Kit) is an experimental WebAssembly standalone
```
jgravelle-google: Why do we need two Wack.cmakes that aren't the same?

src/build.py
```diff
@@ -937,6 +936,16 @@ def LLVM():
   proc.check_call(['ninja', '-v'] + jobs, cwd=LLVM_OUT_DIR, env=cc_env)
   proc.check_call(['ninja', 'install'] + jobs, cwd=LLVM_OUT_DIR, env=cc_env)
   CopyLLVMTools(LLVM_OUT_DIR)
+  install_bin = os.path.join(INSTALL_DIR, 'bin')
+  for target in ('clang', 'clang++'):
+    link = os.path.join(install_bin, 'wasm32-' + target)
+    try:
```
sbc100: Why not just say IsWindows here?

src/build.py
```diff
@@ -1259,6 +1268,7 @@ def LibCXX():
   command = [PREBUILT_CMAKE_BIN, '-G', 'Ninja', os.path.join(LIBCXX_SRC_DIR),
              '-DCMAKE_CXX_COMPILER_WORKS=ON',
              '-DCMAKE_C_COMPILER_WORKS=ON',
+             '-DCMAKE_EXE_LINKER_FLAGS=-nostdlib++',
```
sbc100: I understand why this line should fix the current waterfall failures.
What is really don't understand is how it ever worked without this line?
I can repro this locally not BTW (the failure) while jacob can't?
Also, shouldn't this be the default for the libcxx build?   libcxx should not expect an existing libc++ library to exist on the system when it builds should it?

sbc100:
I'd really like to call this `wasm-clang` rather than wasm32-clang is possible.  I have an llvm-side change that adds this as an alias so the two should be the same.

src/build.py
dschuff: Done; yeah this logic was leftover from a different attempt.

src/build.py
dschuff: Yeah I don't understand why it worked either. And yes you'd think that libcxx should not necessarily expect an existing stdlib; although on non-Apple platforms there always is one, e.g. gcc's libstdc++.

src/cmake/Modules/Platform/Wack.cmake
dschuff: This one is the "platform file" which is in a magic special location in a path in the CMAKE_MODULE_PATH set up by the other file. It gets automatically executed when you set `CMAKE_SYSTEM_NAME` in the main file. I was just going to put all the content in here, but I found that the stuff in the other file had to go before this file is invoked, because it affects other CMake-provided files that get executed before this file does.

src/build.py
dschuff: (btw what version of CMake are you both using?)

jgravelle-google: I build almost everything with `build.py`, so the prebuilt cmake (3.7.2)

sbc100: I'm using the waterfall one too.

sbc100: This line doesn't seem related to the rest of the change anymore, and the immediate issue is fixed.  Can we drop this line for now?

src/build.py
```diff
@@ -937,6 +936,16 @@ def LLVM():
...
+  install_bin = os.path.join(INSTALL_DIR, 'bin')
+  for target in ('clang', 'clang++'):
+    link = os.path.join(install_bin, 'wasm32-' + target)
+    if not IsWindows():
+      os.symlink(Executable(link), Executable(target))
+    else:
+      # Windows has no symlinks (at least not from python). Also clang won't
+      # work as a native compiler anyway, so just install it as wasm32-clang
+      os.rename(Executable(os.path.join(install_bin, target)), Executable(link))
+
```
sbc100: remove extra newline?

src/Wack.cmake
```diff
@@ -0,0 +1,37 @@
...
+
+set(WASM_SDKROOT ${CMAKE_CURRENT_LIST_DIR})
+list(APPEND CMAKE_MODULE_PATH "${WASM_SDKROOT}/cmake/Modules")
+
+# This has to go before we set CMAKE_SYSTEM_NAME because the default c compiler
+# gets set before the platform file is included
+if (CMAKE_HOST_WIN32)
+        set(EXE_SUFFIX ".exe")
+else()
+        set(EXE_SUFFIX "")
```
sbc100: Two space indentation.

dschuff:
Ping! any more thoughts on this? I'm hoping to get a fix on top of this that will finally make the windows bot green again.

PR #342, opened 2018-03-15
Go back to using d8 instead of node for emscripten tests

dschuff:
It's usable again after https://chromium-review.googlesource.com/c/964221/


dschuff:
Yeah, seems reasonable, I'll add D8_BIN

PR #343, opened 2018-03-16
Add em_pre.js, use it to add waterfall-specific overrides for Emscripten-built torture tests

jgravelle-google:


dschuff:
Oh, I'm thinking of https://github.com/kripken/emscripten/pull/5948.
I'm ok with this as a workaround for now, but this seems like a bug we should fix in d8.

dschuff:
Btw is this a new bug? Basically any throw from any wasm test is going to be in a promise context because everything is kicked off from the `then` of instantiation. I guess emscripten wraps a giant try around main though, so it's never been an issue? But we've had plenty of aborts, so I don't understand how this ever worked otherwise.

jgravelle-google:
One we instantiate and start calling `main` we're fine.
This problem isn't aborts in general (they get turned into exceptions that get caught by `callMain` and then passed to `quit`), but with failing to instantiate a module. We don't usually make wasm binaries that don't validate, and earlier steps tend to catch it when we do (e.g. compile). So this feels fairly novel.

dschuff:
I see. I also tested SM and it seems to also ignore unhandled promise rejections. So it's node that's different in this case, but this behavior seems like what we want for these kind of standalone torture tests.

src/em_pre.js
```diff
@@ -0,0 +1,10 @@
+var Module = {}
+
+Module['onAbort'] = function(reason) {
+  // d8 does not exit with status 1 when a promise throws. Emscripten calls
```
dschuff: might want to say  'JS shells' instead of just 'd8' and maybe also mention both throws and rejections (since they are equivalent for most purposes)

PR #345, opened 2018-03-21
Create a clang symlink named for the wasm target

dschuff:
When clang is invoked as `${triple}-clang`, it automatically infers its default target triple and does not need the `--target` flag to select the triple. 
This PR changes the wack installation and CMake files to use `wasm32-clang` instead of just 'clang'.


dschuff:
Forked from #341

dschuff:
@sbc100 did you say you had an LLVM change to make `wasm` an alias in the triple for `wasm32`?

sbc100:
I was working on it but I didn't land it yet.  I'm OK with this landing as is and we can look at using `wasm` rather than `wasm32` as a separate issue.

PR #346, opened 2018-03-22
Use cmake linker flags to get correct CMake feature tests

dschuff:
compiler-rt is especially ugly because without `-nostdlib` it fails because compiler-rt isn't there, but without libc, `_start` is undefined.


sbc100:
I guess this can be closes?

dschuff:

PR #351, opened 2018-05-04
Remove final argument from setup_toolchain.py call

sbc100:
This argument is not longer needed:
  https://chromium-review.googlesource.com/1015860


dschuff:
Thanks, sorry I missed this yesterday.

PR #359, opened 2018-06-14
LLD is now the default to invert the environment settings

sbc100:
To match https://github.com/kripken/emscripten/pull/6697

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1386,7 +1386,7 @@ def CompileLLVMTortureBinaryen(name, em_config, outdir, fails, opt, lld):
   else:
     config = 'binaryen'
   try:
-    if not lld
+    if not lld:
       os.environ['EMCC_EXPERIMENTAL_USE_LLD'] = '0'
     unexpected_result_count = compile_torture_tests.run(
         cc=cc, cxx=cxx, testsuite=GCC_TEST_DIR,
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -1385,14 +1385,21 @@ def CompileLLVMTortureBinaryen(name, em_config, outdir, fails, opt, lld):
...
-  unexpected_result_count = compile_torture_tests.run(
-      cc=cc, cxx=cxx, testsuite=GCC_TEST_DIR,
-      sysroot_dir=INSTALL_SYSROOT,
-      fails=fails,
-      exclusions=LLVM_TORTURE_EXCLUSIONS,
-      out=outdir,
-      config=config,
-      opt=opt)
+  try:
+    if not lld
```
jgravelle-google: missing `:`

src/build.py
```diff
@@ -1385,14 +1385,21 @@ def CompileLLVMTortureBinaryen(name, em_config, outdir, fails, opt, lld):
...
+        cc=cc, cxx=cxx, testsuite=GCC_TEST_DIR,
+        sysroot_dir=INSTALL_SYSROOT,
+        fails=fails,
+        exclusions=LLVM_TORTURE_EXCLUSIONS,
+        out=outdir,
+        config=config,
+        opt=opt)
+  finally:
+    if not lld:
+      del os.environ['EMCC_EXPERIMENTAL_USE_LLD']
```
jgravelle-google: Not a fan of modifying os.environ in general
And in theory this can cause weirdness because it doesn't perfectly clean up (if the var was set before, now it's unset)
But in practice we control all the paths through here
And we'll remove the flag in a week or two tops, so this is as clean as it needs to be

sbc100: Yeah, I figure the same.  Plus I'm only moving existing code

PR #361, opened 2018-06-15
Try a fix for mac builder

dschuff:
Remove the O_NONBLOCK flag from the stdout fd. Taken from a link in a comment on crbug.com/829034

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -289,11 +289,10 @@ BUILDBOT_BUILDERNAME = os.environ.get('BUILDBOT_BUILDERNAME', None)
 
 if IsMac() and IsBuildbot():
   # Experimental temp fix for crbug.com/829034 stdout write sometimes fails
-  import fcntl
-  from os import O_NONBLOCK
+  from fcntl import fcntl, F_GETFL, F_SETFL
   fd = sys.stdout.fileno()
-  flags = fcntl.fcntl(fd, fcntl.F_GETFL)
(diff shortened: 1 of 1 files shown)


dschuff:
I have no idea whether or not this will work, but it's worth a shot.

src/build.py
```diff
@@ -287,6 +287,13 @@ def JavaBinDir():
 BUILDBOT_BUILDNUMBER = os.environ.get('BUILDBOT_BUILDNUMBER', None)
 BUILDBOT_BUILDERNAME = os.environ.get('BUILDBOT_BUILDERNAME', None)
 
+if IsMac() and IsBuildbot():
+  # Experimental temp fix for crbug.com/829034 stdout write sometimes fails
+  import fcntl
+  from os import O_NONBLOCK
```
jgravelle-google: Nit: Why not `os.O_NONBLOCK`?

src/build.py
```diff
@@ -287,6 +287,13 @@ def JavaBinDir():
...
+  fd = sys.stdout.fileno()
+  flags = fcntl.fcntl(fd, fcntl.F_GETFL)
+  fcntl.fcntl(fd, fcntl.F_SETFL, flags & ~O_NONBLOCK)
```
jgravelle-google: Nit: See it might make more sense to say `from fcntl import (fcntl, F_GETFL, F_SETFL)` because that's lot of `fcntl` in a short space

PR #362, opened 2018-06-18
Remove s2wasm testing from the waterfall

sbc100:
This is a fairly large change since many part of the waterfall
were using s2wasm or its output.

We now only compiler the toture tests once, to wasm object files.
However we link them twice, once with -nostdlib and once without.
This is a because we can't run binaries linked with musl in the
spec interpreter.


dschuff:
So IIUC this means we no longer have the 'bare' vs 'bare-musl' distinction when running with d8; the d8 tests always link with musl, and the spec interpreter (and binaryen interpreter?) link with nostdlib?

src/build.py
```diff
@@ -1565,11 +1533,9 @@ def Test(self):
 
 def GetTortureDir(name, opt):
   dirs = {
-      's': os.path.join(TORTURE_S_OUT_DIR, opt),
       'o': os.path.join(TORTURE_O_OUT_DIR, opt),
       'asm2wasm': os.path.join(ASM2WASM_TORTURE_OUT_DIR, opt),
-      'emwasm': os.path.join(EMWASM_TORTURE_OUT_DIR, opt),
-      'emwasm-lld': os.path.join(EMWASM_LLD_TORTURE_OUT_DIR, opt)
+      'emwasm': os.path.join(EMWASM_TORTURE_OUT_DIR, opt)
```
jgravelle-google: Nittiest of nits: can leave the trailing `,` at the end of the line, which reduces line churn in diffs when adding elements, and makes it iotally easier to add or reorder elements.

Not all languages allow this, but python does, and it's pretty nice

kripken:
Looks like a lot of known failures went away here, just curious, what are the causes?

sbc100:
Basically the spec interpreter is only setup to run the nostdlib tests.  Previously these were generated directly by s2wasm + wat2wasm.  Now the same set of the tests are instead build with `clang -nostdlib`.  

I believe that d8 was already only running the libc-linked wasm filed

PR #363, opened 2018-06-19
Remove wasm-shell testing of torture tests

sbc100:
This is part of the move to remove s2wasm.  When we do this
we will only producing wasm binaries from the toolchain and
not wast/wat files.

wasm-shell only supports running on the .wast files (it designed
to run the spec tests).  (It you pass it wasm file it actually just
returns 1 since wasm files are all empty strings in C!)

The wasm-shell tests have been failing for a while but we have
the in warn only mode, which is another reason I don't think we
care about coverage here.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -254,9 +254,6 @@ S2WASM_KNOWN_TORTURE_FAILURES = [os.path.join(BINARYEN_SRC_DIR, 'test',
                                               's2wasm_' + IT_IS_KNOWN)]
 LLD_KNOWN_TORTURE_FAILURES = [os.path.join(SCRIPT_DIR, 'test',
                               'lld_' + IT_IS_KNOWN)]
-BINARYEN_SHELL_KNOWN_TORTURE_FAILURES = [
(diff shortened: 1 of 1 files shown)


sbc100:
Presumably wasm-shell is design to run the spec tests..  aren't those sufficient to consider it tested?

jgravelle-google:
> it would be pretty easy to make it work on binaries too

Looking at the code it's pretty non-standard for how binaryen handles wasm input. Instead of using `ModuleReader` it uses `SExpressionParser`, then iterates over the `Element`s to pick out the test assertions and run those.
Which is to say it's running what it thinks are spec tests exclusively. It wouldn't be impossible to extend it to binaries, but it'd be changing what the tool is presumably for.

dschuff:
Oh? but we don't put assertions in s2wasm output, so for non-spec tests it only uses the flag that sets the entry point.

dschuff:
I guess in a binary mode, we could just only support a single entry point.

PR #369, opened 2018-06-25
Improve output of test execution when no know failures file is provided

sbc100:
This path wasn't being tested it seemed so had accumulated
some bugs.

### DIFF
--- a/src/testing.py
+++ b/src/testing.py
@@ -247,15 +247,15 @@ def execute(tester, inputs, fails, exclusions=None, attributes=None):
     sys.stdout.write(str(result) + '\n\n')
 
   if not fails:
-    sys.stdout.write('\n'.join([
-                 'Ran %s tests.' % len(results),
-                 'Got %s successes.' % len(successes),
-                 'Got %s failures.' % len(failures)]))
+    sys.stdout.write(
+        '\n'.join(['Ran %s tests.' % len(results),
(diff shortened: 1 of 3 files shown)


src/testing.py
```diff
@@ -241,27 +237,40 @@ def execute(tester, inputs, fails, exclusions=None, attributes=None):
...
-    return failures
+    sys.stdout.write('\n'.join([
+                 'Ran %s tests.' % len(results),
+                 'Got %s successes.' % len(successes),
+                 'Got %s failures.' % len(failures)]))
+    if failures:
+      sys.stdout.write('\nUnexpected failures:\n')
+      for f in failures:
+        sys.stdout.write('\t%s\n' % f.test)
+    return  len(failures)
```
jgravelle-google: nit: extra space between `return` and `len`

src/testing.py
```diff
@@ -241,27 +237,40 @@ def execute(tester, inputs, fails, exclusions=None, attributes=None):
...
   failures = [r for r in results if not r]
+
+  sys.stdout.write('\nResults:\n')
+  for result in results:
+    sys.stdout.write(str(result) + '\n\n')
+
   if not fails:
-    return failures
+    sys.stdout.write('\n'.join([
+                 'Ran %s tests.' % len(results),
```
jgravelle-google: nit: indentation here doesn't fit the previous line

PR #371, opened 2018-06-25
Add validate of llvm object files

sbc100:
The object files produced by llvm should be valid wasm files
so its good to test this too.


src/test/validate_known_gcc_test_failures.txt
```diff
@@ -0,0 +1,12 @@
+# Expected validation failures in output of llvm backend.
+# Failure here mean the llvm is producing bad wasm. This should never happen :(
```
jgravelle-google: nit: "mean the llvm" -> "means that llvm"?

PR #374, opened 2018-06-28
Update expectations after upstream clang change

sbc100:


jgravelle-google:
Fix should be in https://github.com/llvm-mirror/clang/commit/ff6b0ecacdfc5a066f35b140db832b02082f9731, so we probably don't need this

PR #379, opened 2018-07-09
Add remaining LUCI configs

rryk:


rryk:
I'd like to add @mi-ac to this review, but for some reason it does not show up in the list of the reviewers. Does he also need to be part of the WebAssembly project to be able to review this?

dschuff:
I don't know about the reviewer list but that's a plausible explanation. It looks like everyone I see in that list is in the org. I'm happy to wait for @mi-ac 's review before landing though.

cr-buildbucket.cfg
```diff
@@ -0,0 +1,56 @@
...
+    builders {
+      name: "linux"
+      dimensions: "cpu:x86-64"
+      dimensions: "os:Ubuntu-14.04"
+    }
+
+    builders {
+      name: "windows"
+      dimensions: "cpu:x86-64"
+      dimensions: "os:Windows"
```
mi-ac: Maybe specify OS version here, too?

cr-buildbucket.cfg
```diff
@@ -0,0 +1,56 @@
...
+      build_numbers: YES
+    }
+
+    builders {
+      name: "linux"
+      dimensions: "cpu:x86-64"
+      dimensions: "os:Ubuntu-14.04"
+    }
+
+    builders {
```
mi-ac: Pool dimension is missing. Please add one.

mi-ac: Never mind. Realized again that pool will default to bucket name...

PR #381, opened 2018-07-11
Remove known validation failures for llvm output

sbc100:
The signature mismatch problem with prototype-less functions
has now been fixed.


sbc100:
Perhaps?   I sort of like having the validation as a separate pass.

PR #383, opened 2018-07-11
Remove OS version specifications and add mastername property

rryk:
TBR @mi-ac.


rryk:
Addressed in a follow-up: https://github.com/WebAssembly/waterfall/pull/388

PR #402, opened 2018-08-18
Build the native asmjs optimizer in a separate step

dschuff:
The way embuilder builds the native optimizer with cmake doesn't work on the waterfall on Windows; and in any case it's better to force the optimizer binary anyway, to avoid building it on-demand.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1125,7 +1125,6 @@ def Emscripten():
                   # Ignore the big git blob so it doesn't get archived.
                   ignore=shutil.ignore_patterns('.git'))
 
-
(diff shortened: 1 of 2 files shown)


sbc100:
Shouldn't we fix the embuilder build?

dschuff:
It's really more of an issue with the waterfall. If you have MSVC properly installed on the system, with the environment variables in your environment, then things will just work. But the Chrome bots don't have the toolchain installed in the usual way, so we have to jump through hoops to build anything. We could add the environment to the embuilder build, but I think putting the optimizer path in the emscripten config file (as we do for LLVM) makes sense for this packaging anyway.

PR #403, opened 2018-08-22
Use gclient-managed hermetic mac toolchain on mac bots

dschuff:
Experimental fix to bring up the mac LUCI bot

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -288,6 +288,15 @@ if IsMac() and IsBuildbot():
   flags = fcntl(fd, F_GETFL)
   fcntl(fd, F_SETFL, flags & ~os.O_NONBLOCK)
 
+  # Use gclient-managed hermetic toolchain on mac LUCI bots.
+  # TODO: Convert bot build to recipes, and this should no longer be necessary.
+  # Force gclient to install the hermetic mac toolchain
+  os.environ['FORCE_MAC_TOOLCHAIN'] = 1
+  # The V8 build_overrides/build.gni specifically forces use of the system
+  # Xcode, so override the override.
+  os.environ['DEVELOPER_DIR'] = os.path.join(V8_SRC_DIR,
+                                             'build', 'mac_files', 'Xcode.app')
+
 # Pin the GCC revision so that new torture tests don't break the bot. This
 # should be manually updated when convenient.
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -288,6 +288,15 @@ def IsBuildbot():
...
   fcntl(fd, F_SETFL, flags & ~os.O_NONBLOCK)
 
+  # Use gclient-managed hermetic toolchain on mac LUCI bots.
+  # TODO: Convert bot build to recipes, and this should no longer be necessary.
+  # Force gclient to install the hermetic mac toolchain
+  os.environ['FORCE_MAC_TOOLCHAIN'] = 1
+  # The V8 build_overrides/build.gni specifically forces use of the system
+  # Xcode, so override the override.
+  os.environ['DEVELOPER_DIR'] = os.path.join(V8_SRC_DIR,
+                                             'build', 'mac_files', 'Xcode.app')
```
sbc100: This feels like it should be part of V8 build only, no?   Or does it effect other builds too?

dschuff: The intent was to make it affect all the builds, since in general it seems best to use the same toolchain everywhere. I still don't understand why the non-V8 builds are working when the V8 build is not, though (maybe the failing check is overly-specific? Don't know).  Eventually I think we want to move this bot fully over to LUCI-style recipes (and my understanding is that LUCI bots have a different way of managing the toolchain than I'm using here). So this is supposed to be a temporary fix.

In any case I also found that none of this is actually running anyway because LUCI builders don't set `BUILDBOT_BUILDNUMBER` along with the other environment vars. So I also need to figure out how to get the LUCI build number if we want to archive any builds.

sbc100:
Oops:  

```
__CF_USER_TEXT_ENCODING: 0x1F4:0x0:0x0
  File "/b/build/slave/mac/build/src/src/build.py", line 294, in <module>
    os.environ['FORCE_MAC_TOOLCHAIN'] = 1
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 473, in __setitem__
    putenv(key, item)
TypeError: must be string, not int
step returned non-zero exit code: 1
```

PR #407, opened 2018-08-24
Copy concrt140.dll along with the rest of the VS runtime

dschuff:
LLD needs this dll, but it isn't copied by Chromium's script.

### DIFF
--- a/src/host_toolchains.py
+++ b/src/host_toolchains.py
@@ -15,8 +15,10 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import glob
 import json
 import os
+import shutil
 
 import file_util
 import proc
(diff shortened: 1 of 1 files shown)


src/host_toolchains.py
```diff
@@ -94,7 +106,10 @@ def SetUpVSEnv(outdir):
 def CopyDlls(dir, configuration):
   """Copy MSVS Runtime dlls into a build directory"""
   proc.check_call([VS_TOOLCHAIN, 'copy_dlls', dir, configuration, 'x64'])
-
+  # LLD needs also concrt140.dll, which the Chromium copy_dlls doesn't include.
+  for dll in glob.glob(os.path.join(GetRuntimeDir(), 'concrt140*.dll')):
+    print 'Copying %s to %s' % (dll, dir)
+    shutil.copy2(dll, dir)
```
sbc100: I wonder why?  Presumably chrome is using lld so they should need all the same dlls?

dschuff:
Chrome may or may not be using lld (clang works with gnu ld, gold and link.exe) but this dll-copying is for distribution, i.e. running on a machine other than a build machine. So Chrome itself doesn't necessarily need the same dlls.

src/host_toolchains.py
dschuff: Chrome may or may not be using lld to link (clang works with gnu ld, gold and link.exe) but this dll-copying is for distribution, i.e. running on a machine other than a build machine. So Chrome itself doesn't necessarily need the same dlls.

PR #410, opened 2018-09-20
Update node version to 8.12.0 and switch to windows zip package

dschuff:
The bare Windows binary used previously didn't support WebAssembly. That's now required in order to build struct_info.json for the wasm backend (since that configuration can't build asmjs).

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -164,7 +164,7 @@ def WindowsFSEscape(path):
 def NodePlatformName():
   return {'darwin': 'darwin-x64',
           'linux2': 'linux-x64',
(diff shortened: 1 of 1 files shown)


dschuff:
I have no idea now this ever worked before.

sbc100:
Any reason you picked 8.12.0?  Is there some kind of LTS for node?

dschuff:
8.12 is the current LTS. I would actually not object to going even newer here if there's a good reason. One good reason might be to get node's implementation of the worker API to allow testing of pthread code outside the browser.

PR #414, opened 2018-10-02
Remove two known failures from spec_known_gcc_test_failures.txt

sbc100:
Since the default for lld changed to --no-export-dynamic we
not longer export -fvisibility=default symbols.  These two
tests started passing because in the O2 case the operator new
is no longer included in the link.

It seems that operator new always visibility=default so is one
of the few symbols used in the torture tests that is effected.


jgravelle-google:
lgtm

PR #417, opened 2018-10-03
Flush stdout after multiprocessing pool join

dschuff:
Mac bot builds currently always fail with EAGAIN on stdout.write calls on line
247 below. One possible explanation might be a full buffer somewhere underneath,
so try flushing stdout after joining the pool and before trying to write
anything else.


dschuff:
Indeed. That's why it's likely wrong :woman_shrugging:

PR #418, opened 2018-10-30
WebAssembly is no longer and experimental target

sbc100:


src/build.py
```diff
@@ -900,8 +900,7 @@ def LLVM():
              '-DLLVM_TOOL_LTO_BUILD=OFF',
              '-DLLVM_INSTALL_TOOLCHAIN_ONLY=ON',
              '-DLLVM_ENABLE_ASSERTIONS=ON',
-             '-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly',
-             '-DLLVM_TARGETS_TO_BUILD=X86']
+             '-DLLVM_TARGETS_TO_BUILD=X86,WebAssembly']
```
aheejin: Don't we need "", like `"X86,WebAssembly"`?

dschuff: Not sure, but I'm guessing no because we don't use quotes in any other CMake flags?

aheejin: At least it doesn't work on my shell without quotes. And [LLVM CMake document](https://llvm.org/docs/CMake.html#llvm-specific-variables) also says

> LLVM_TARGETS_TO_BUILD:STRING
> Semicolon-separated list of targets to build, or all for building _all_ targets. Case-sensitive. Defaults to _all_. Example: `-DLLVM_TARGETS_TO_BUILD="X86;PowerPC"`.

PR #424, opened 2018-11-07
Update infra configs

rryk:


rryk:
Looks like this also pulled my earlier change for specific OS versions, which is useful too. I've changed the pull request title to a more generic one capturing both of these commits.

PR #425, opened 2018-11-20
Remove expected failures due to __builtin_constant_p

sbc100:
I seems that references to this symbol are not longer present
in clangs -O2, so these tests now pass.


jgravelle-google:
Of course this might immediately re-red us: https://wasm-stat.us/builders/linux/builds/37371

sbc100:
Hmm, how did that happen?  maybe an llvm change got reverted?  I will revert his.

PR #428, opened 2018-11-30
Explicitly tell CMake python's location

tlively:
Introduces `CMakeCommand()`, which assembles all the common parts of a
CMake command, including the specification of the python
location. This adds the result of `OverrideCMakeCompiler()` and
`host_toolchains.CMakeLauncherFlags()` to CMake commands that did not
already have them, but hopefully that doesn't break anything.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -825,12 +825,25 @@ def OverrideCMakeCompiler():
           '-DCMAKE_CXX_COMPILER=' + CXX]
 
 
-def CMakeCommand(args):
+def CMakeCommandBase(args):
   command = [PREBUILT_CMAKE_BIN, '-G', 'Ninja'] + args
   # Python's location could change, so always update CMake's cache
-  command.append('-DPYTHON_EXECUTABLE=%s' % proc.Which('python', os.getcwd()))
+  command.append('-DPYTHON_EXECUTABLE=%s' % sys.executable)
+  command.append('-DCMAKE_EXPORT_COMPILE_COMMANDS=ON')
+  command.append('-DCMAKE_BUILD_TYPE=Release')
+  return command
+
+
+def CMakeCommandNative(args):
+  command = CMakeCommandBase(args)
   command.extend(OverrideCMakeCompiler())
   command.extend(host_toolchains.CMakeLauncherFlags())
+  return command
+
+
+def CMakeCommandWack(args):
+  command = CMakeCommandBase(args)
+  command.append('-DCMAKE_TOOLCHAIN_FILE=%s' % CMAKE_TOOLCHAIN_FILE)
   return command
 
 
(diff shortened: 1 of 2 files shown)


src/build.py
```diff
@@ -825,6 +825,15 @@ def OverrideCMakeCompiler():
           '-DCMAKE_CXX_COMPILER=' + CXX]
 
 
+def CMakeCommand(args):
+  command = [PREBUILT_CMAKE_BIN, '-G', 'Ninja'] + args
+  # Python's location could change, so always update CMake's cache
+  command.append('-DPYTHON_EXECUTABLE=%s' % proc.Which('python', os.getcwd()))
```
dschuff: This should probably just be `sys.executable` so everything is always guaranteed to use the same python.

src/build.py
```diff
@@ -825,6 +825,15 @@ def OverrideCMakeCompiler():
...
+  command.extend(OverrideCMakeCompiler())
```
dschuff: So there are really 2 different types of builds we do: native builds, and cross-builds using the wasm compiler. It looks like Python is really the only thing that they have in common (edit: also I guess `CMAKE_BUILD_TYPE`). `OverrideCMakeCompiler` only applies to native builds (for wasm builds `CC` is set by the toolchain file). Also Goma only works for native builds (that's what the `CMakeLauncherFlags` is for). 
So maybe it makes sense to just have 2 different `CMakeCommand` functions, i.e. `CMakeCommandNative` and `CMakeCommandWasmClang` or (`CMakeCommandWack`). Then `OverrideCMakeCompiler`, `CMakeLauncherFlags`, and `CMAKE_INSTALL_PREFIX` can go in the native one, and `CMAKE_TOOLCHAIN_FILE` can go in the wasm one.
Come to think of it, I'm not even sure Python is used by any of the wasm builds, but I guess it wouldn't hurt to keep it in both just to be safe.

src/build.py
```diff
@@ -864,26 +873,25 @@ def LLVM():
...
-             # Our mac bot's toolchain's ld64 is too old for trunk libLTO.
-             '-DLLVM_TOOL_LTO_BUILD=OFF',
-             '-DLLVM_INSTALL_TOOLCHAIN_ONLY=ON',
-             '-DLLVM_ENABLE_ASSERTIONS=ON',
-             '-DLLVM_TARGETS_TO_BUILD=X86;WebAssembly']
-
-  command.extend(OverrideCMakeCompiler())
+  command = CMakeCommand([
+      LLVM_SRC_DIR,
+      '-DCMAKE_EXPORT_COMPILE_COMMANDS=ON',
```
dschuff: Might as well move this into `CMakeCommand` too actually. If it's useful anywhere (which I'm not 100% sure it is, but :woman_shrugging:) then it's useful in all/any of the CMake uses.

src/build.py
```diff
@@ -864,26 +886,22 @@ def LLVM():
...
-             # Our mac bot's toolchain's ld64 is too old for trunk libLTO.
-             '-DLLVM_TOOL_LTO_BUILD=OFF',
-             '-DLLVM_INSTALL_TOOLCHAIN_ONLY=ON',
-             '-DLLVM_ENABLE_ASSERTIONS=ON',
-             '-DLLVM_TARGETS_TO_BUILD=X86;WebAssembly']
-
-  command.extend(OverrideCMakeCompiler())
+  command = CMakeCommandNative([
+      LLVM_SRC_DIR,
+       '-DCMAKE_INSTALL_PREFIX=' + INSTALL_DIR,
```
dschuff: I think `CMAKE_INSTALL_PREFIX` can go into `CMakeCommandNative` too. The emscripten optimizer build below doesn't do an install, so it doesn't matter whether it gets set.

PR #429, opened 2018-12-03
Restore fastcomp's install prefix

tlively:


tlively:
This won't work. It will get overridden in CMakeCommandNative. Brb.

sbc100:
Although won't this arg now appear twice for fastcomp?

tlively:
Yes, but I think the later one wins.

tlively:
This patch gets me fastcomp's clang locally.

PR #430, opened 2018-12-04
Update test expectations

sbc100:
For opt__pr15551.C it looks like the call to operator new is no
longer elided.


tlively:
Thanks for keeping on top of these!

PR #431, opened 2018-12-11
Add landmine-like system for triggering clobber builds

sbc100:


sbc100:
Actually tested locally now

PR #434, opened 2018-12-14
Only clobber work directory if we are on a bot

sbc100:


src/build.py
```diff
@@ -726,6 +726,10 @@ def get_rev(rev_name):
 
 
 def Clobber():
+  # Don't ever clobber non-bot (local) work directories
+  if not buildbot.IsBot():
```
dschuff: Should this just be rolled into `ShouldClobber()`?

sbc100: Or do you mean this whole function?  If thats what you meant I'm not sure it really matters.

dschuff: Oh I see what you mean. buildbot.ShouldClobber just checks the buildbot-controlled clobbering. This change is fine; separately we should just remove ShouldClobber since we don't have buildbot anymore.

PR #435, opened 2018-12-14
Use print instead of sys.stdout.write in testing.py

dschuff:
Yet another experiment in fixing crbug.com/829034
Based on the observation that testing.py-run tests are the ones failing, and testing.py uses sys.stdout.write instead of print; try just not doing that. Just in case the print function does something interesting with i/o buffering or something.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -272,7 +272,8 @@ if IsMac():
   from fcntl import fcntl, F_GETFL, F_SETFL
   fd = sys.stdout.fileno()
   flags = fcntl(fd, F_GETFL)
-  fcntl(fd, F_SETFL, flags & ~os.O_NONBLOCK)
+  print 'Flags: %s, nonblock: ' % (hex(flags), flags & os.O_NONBLOCK)
+  #fcntl(fd, F_SETFL, flags & ~os.O_NONBLOCK)
 
 # Pin the GCC revision so that new torture tests don't break the bot. This
 # should be manually updated when convenient.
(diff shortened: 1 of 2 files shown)


src/build.py
```diff
@@ -267,12 +267,13 @@ def JavaBinDir():
 NPROC = multiprocessing.cpu_count()
 
 
-if IsMac():
+if IsMac() or True:
```
aheejin: Does `or True` mean we are gonna run this in every platform?

src/build.py
```diff
@@ -267,12 +267,13 @@ def JavaBinDir():
...
   # Experimental temp fix for crbug.com/829034 stdout write sometimes fails
-  from fcntl import fcntl, F_GETFL, F_SETFL
+  from fcntl import fcntl, F_GETFL
   fd = sys.stdout.fileno()
   flags = fcntl(fd, F_GETFL)
-  fcntl(fd, F_SETFL, flags & ~os.O_NONBLOCK)
+  print 'Flags: %s, nonblock: %s' % (hex(flags), flags & os.O_NONBLOCK)
```
aheejin: - You used Python 3 style `print` in `testing.py`... did you mean to do it here too?
- Not sure what that exactly means, but code in the left side is `flag & ~os.O_NONBLOCK` and this is `flags & ~os.O_NENBLOCK`, with missing `~`. Is this intentional?

src/build.py
dschuff: Oops, that was for local debugging.

src/build.py
dschuff: * No I just converted that one file because with the print function I could to a search and replace for sys.stdout.write and put print instead. But I didn't change the rest of the files. If this PR doesn't help I'll probably just revert it. Otherwise I'll go in and clean it up
* Previously the code was clearing the nonblock flag, and now the code is just printing the value

PR #436, opened 2018-12-18
Revert "Use print instead of sys.stdout.write in testing.py (#435)"

dschuff:
This reverts commit f060c16cfd6d1fa128d60af551f2885f6aa2f7ea.


dschuff:
alas, it didn't help.

PR #440, opened 2018-12-18
Replace multiprocessing pool with custom parallel test runner in testing.py

dschuff:
This avoids printing to stdout from the subprocesses, in hopes of working around the EAGAIN errors on mac.


dschuff:
Gonna land this to see if it actually works; if so we'll do a proper review and also make it print the command lines in another way.

PR #441, opened 2018-12-19
Do not print list of binaries on Mac

rryk:


dschuff:
Actually, can you fix the flake8 error from ci:
```
./src/build.py:1339:5: E121 continuation line under-indented for hanging indent
The command "flake8" exited with 1.
```

PR #442, opened 2018-12-19
Increase timeout for wasm builds

rryk:


rryk:
Looks like 4 hours was not enough: https://ci.chromium.org/p/wasm/builders/luci.wasm.ci/mac/4042. I have re-triggered this build manually (via led) with a 10h timeout to see how much time we actually need: https://ci.chromium.org/swarming/task/41e3f48dddf55d10.

PR #443, opened 2019-01-14
Remove frame-address.c from expected failures

aheejin:
It used to fail when any optimization flag was on and the reason was
RegStackify's stack pointer store check was incorrect. This test is now
passing after the LLVM fix (https://reviews.llvm.org/rL350906).

In detail: In `check_fa_mid`, the original code order was
```
  ...
  i32.call  check_fa_work@FUNCTION
  global.set  __stack_pointer@GLOBAL
  ...
```

But the previous code incorrectly reordered the call to `check_fa_work`
after `global.set` while stackifying registers. `check_fa_work` has
`global.get  __stack_pointer@GLOBAL` instruction itself, resulting in
loading an incorrect value.

### DIFF
--- a/src/test/run_known_gcc_test_failures.txt
+++ b/src/test/run_known_gcc_test_failures.txt
@@ -123,8 +123,6 @@ pr49279.c.js asm2wasm
 # aborts in native clang
 20031003-1.c.js O0 # abort() in emwasm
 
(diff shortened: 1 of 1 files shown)


aheejin:
Ping

PR #446, opened 2019-01-16
Reduce nunber of tests on the mac bot

dschuff:
The bot is currently timing out, so rather than increasing the timeout,
reduce the test matrix


dschuff:
Oops, my failure to properly spell 'number' is now permanently a part of history.

PR #447, opened 2019-01-16
Update LLVM build to use the monorepo

dschuff:
The official LLVM monorepo is now live, so it can be used instead of the individual git mirrors.
This means we can also remove the logic to try to synchronize clang+llvm via SVN.


dschuff:
This will be somewhat disruptive to anyone who is using the waterfall for local development.
As written, this will check out the monorepo into `src/work/llvm-project`. On local non-bot checkouts, it will leave behind `src/work/{llvm,clang,lld,libcxx,libcxxabi,compiler-rt}` but they will not be updated.
@aheejin @tlively @kripken @jgravelle-google @sbc100 @aardappel

src/build.py
```diff
@@ -919,7 +857,9 @@ def LLVM():
       '-DLLVM_TOOL_LTO_BUILD=OFF',
       '-DLLVM_INSTALL_TOOLCHAIN_ONLY=ON',
       '-DLLVM_ENABLE_ASSERTIONS=ON',
-      '-DLLVM_TARGETS_TO_BUILD=X86;WebAssembly'
+      '-DLLVM_TARGETS_TO_BUILD=X86;WebAssembly',
+      '-DLLVM_EXTERNAL_CLANG_SOURCE_DIR=%s' % CLANG_SRC_DIR,
+      '-DLLVM_EXTERNAL_LLD_SOURCE_DIR=%s' % LLD_SRC_DIR
```
sbc100: Why do you need these extra defines?

src/build.py
```diff
@@ -115,6 +114,7 @@
 GITHUB_MIRROR_BASE = GIT_MIRROR_BASE + 'external/github.com/'
 WASM_GIT_BASE = GITHUB_MIRROR_BASE + 'WebAssembly/'
 EMSCRIPTEN_GIT_BASE = GITHUB_MIRROR_BASE + 'kripken/'
+LLVM_GIT_BASE = 'https://github.com/llvm/'
```
sbc100: Should we wait for https://llvm.googlesource.com/ to mirror this repo?

dschuff: Turns out it's already available on https://github.googlesource.com/llvm/llvm-project/, I'll update.

src/build.py
dschuff: Because the monorepo is flat (llvm, clang, and lld are all next to each other, instead of clang being checked out as llvm/tools/clang). So when you build LLVM the source is external rather than in the llvm tree.

src/build.py
binji: Huh, it's been around for a while... how have I never heard of this mirror?

sbc100: Seems like it does: https://github.googlesource.com/WebAssembly/wabt

Looks like most/all of them: https://github.googlesource.com/WebAssembly/

dschuff: Oh, indeed it does! In other news the repo search on that homepage is apparently case sensitive ಠ_ಠ

sbc100:
Well that is a shame... I guess we need to get llvm-project added to llvm.googlesource.com?

sbc100:
Why does github.googlesource.com require auth I wonder?

PR #448, opened 2019-01-21
update location of emscripten repos, which transferred to emscripten-core/

kripken:

PR #449, opened 2019-01-23
Use emscripten from github

kripken:
Followup for #448 - use emscripten repos from github directly. Hopefully will fix https://github.com/WebAssembly/waterfall/pull/448#issuecomment-456619724 

However, I'm not sure why we used chromium.googlesource instead of github?


tlively:
I think we used the googlesource mirrors to avoid depending on github to be up for the bots to run. In other projects it helps reduce load to github as well, but our builders' scale isn't large enough that this matters much.

The googlesource mirrors must not have picked up the repo at the new location, though. We should merge this now to get the waterfall building, but it might be good to switch back once the mirrors are updated.

kripken:
Thanks @tlively. Ok, seeing some more runs meanwhile, it seems this PR can't break things any worse than they already are (it fails in the sync stage 100% of the time), and looks like there are no objections, so landing.

kripken:
The waterfall is green! :)

PR #452, opened 2019-02-11
Update emscripten test mode names after wasm-object-files-by-default landed

kripken:
Also stops testing llvm bitcode files on all the emscripten test suite, like emscripten CI now does (testing focuses on wasm object files, with `other` covering some interesting llvm bitcode linking cases - will be improved as we improve LTO).


kripken:
Note that this is urgent to prevent breakage on the waterfall, after the emscripten repo renamed those test modes.

PR #453, opened 2019-02-12
Cleanup test runner stdout

sbc100:
Add missing newline.
Be little more explicit when printing periodic status.

### DIFF
--- a/src/testing.py
+++ b/src/testing.py
@@ -239,7 +239,7 @@ def execute(tester, inputs, fails, exclusions=None, attributes=None):
   if exclusions:
     input_exclusions = parse_exclude_files(exclusions, None)
     inputs = [i for i in inputs if os.path.basename(i) not in input_exclusions]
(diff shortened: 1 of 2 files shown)


src/testing.py
```diff
@@ -266,9 +266,9 @@ def execute(tester, inputs, fails, exclusions=None, attributes=None):
     sys.stdout.write(
         '\n'.join(['Ran %s tests.' % len(results),
                    'Got %s successes.' % len(successes),
-                   'Got %s failures.' % len(failures)]))
+                   'Got %s failures.' % len(failures)]) + '\n')
     if failures:
-      sys.stdout.write('\nUnexpected failures:\n')
+      sys.stdout.write('Unexpected failures:\n')
```
aheejin: Why delete the newline in the beginning?

sbc100: I added it onto previous print statement instead.. Now each statement is repsonsible for ending its own line :)

PR #458, opened 2019-02-22
Update expectations after lld change

sbc100:
The way in which signature mismatches are handled changed
in https://reviews.llvm.org/D57909.


dschuff:
LGTM

PR #464, opened 2019-03-01
Avoid running d8 tests on windows because of flakiness.

kripken:
The flakiness seems to have gotten worse this week - we've updated the expectations a few times, but really this seems like true flakiness - the same test (not one we've marked earlier this week) started to fail randomly, and we've flickered between red and green builds because of it.

In practice, running the torture tests on one platform is basically enough to test toolchain codegen, and what we were testing on windows was more affected by d8 than anything else.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1489,17 +1489,21 @@ def TestBare():
   common_attrs = ['bare']
   common_attrs += ['win'] if IsWindows() else ['posix']
 
-  for opt in BARE_TEST_OPT_FLAGS:
-    ExecuteLLVMTorture(
-        name='d8',
-        runner=D8_BIN,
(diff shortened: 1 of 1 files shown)


kripken:
Thanks, I forgot the other torture executions.

Excluding on ALL_TESTS would prevent compiling too - which I don't care strongly about. I pushed a commit to just disable the execution in all the torture executions, but happy to exclude even more if you prefer.

dschuff:
oh you're right. that's unfortunate that we have redundant exclusions. But i'd rather keep the compilation step.

once we switch waterfalls we should work toward getting the emscripten test suite or some other suite that does actually execute the code in some way.

PR #469, opened 2019-03-11
Disable printing off archive contents on linux as well as mac

sbc100:
We are seeing consistent errors on the bots:
 `ls: write error: Resource temporarily unavailable`

Seems like it must be a chrome infra issue:
  https://bugs.chromium.org/p/chromium/issues/detail?id=940663


dschuff:
If the experience with the mac bot is any indication, this will just push the problem somewhere else. But hey, it's worth a try.

PR #470, opened 2019-03-12
Run emscripten test suite in single core mode on MacOS

kripken:
This may work around odd hangs on the bots.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1595,11 +1595,24 @@ def ExecuteEmscriptenTestSuite(name, tests, config, outdir, warn_only=False):
 
 
 def TestEmtest():
-  ExecuteEmscriptenTestSuite(
-      'emwasm',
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -1595,11 +1595,24 @@ def ExecuteEmscriptenTestSuite(name, tests, config, outdir, warn_only=False):
 
 
 def TestEmtest():
-  ExecuteEmscriptenTestSuite(
-      'emwasm',
-      ['wasm2', 'other'],
-      EMSCRIPTEN_CONFIG_WASM,
-      EMSCRIPTEN_TEST_OUT_DIR)
+  # Running in parallel hits a problem on the bots where things hang.
+  if IsMac():
```
dschuff: It might be simpler to factor this into context handler.

PR #474, opened 2019-03-14
Remove "nostdlib" build and "spec" interpreter testing

sbc100:
There is little utility in building with nostdlib and running on the
spec interpreter these days.


src/testing.py
```diff
@@ -157,7 +157,7 @@ def __setitem__(self, key, value):
     k = self.canonicalize(key)
     # Support single-insertion only, the intended usage would be a bug if there
     # were multiple insertions of the same key.
-    assert k not in self.arr, 'Double insertion of key %s' % k
```
jgravelle-google: I thought `%s` would `str`-convert args anyway, what's the difference here?

sbc100: This is needed if you want to format a tuple (which this is), otherwise formatting will raise and exception that there are not enough % chars in the string.

PR #476, opened 2019-03-15
Add work_dirs.py to allow overriding the working directories

dschuff:
In preparation for making this script runnable from new bot infrastructure, stop hard-coding the working directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,6 @@
 *~
 *.pyc
-sync/
-build/
-install/
-test-out/
+src/work
 debian/wasm-toolchain.substvars
 debian/wasm-toolchain
 debian/wasm-toolchain.postinst.debhelper
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 *~
 *.pyc
-work/
+sync/
+build/
+install/
+test-out/
 debian/wasm-toolchain.substvars
 debian/wasm-toolchain
 debian/wasm-toolchain.postinst.debhelper
--- a/src/build.py
+++ b/src/build.py
@@ -78,7 +78,6 @@ JSVU_OUT_DIR = os.path.expanduser(os.path.join('~', '.jsvu'))
 
 
 def GetInstallDir(*args):
-  #args = [] if args is None else list(args)
   return os.path.join(work_dirs.GetInstall(), *args)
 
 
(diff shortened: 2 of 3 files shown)


dschuff:
This is still WIP, but moving the default build and install directories might disrupt local testing workflows, so this is a pre-heads up for @aheejin @jgravelle-google @kripken  @sbc @tlively @aardappel 

It would be nice (cleaner) to also move the default sync directory, but that has the potential to be *really* disruptive, since we tend to keep our working source repos for some projects in there.

dschuff:
Actually, it's probably not particularly any cleaner to change the defaults (everything under 'work/') to something else inside the waterfall root directory. It's just sufficient to allow flags to control those directories explicitly so they can be made outside the waterfall directory if desired. I'll revert the defaults.

src/build.py
```diff
@@ -784,8 +780,9 @@ def BuildEnv(build_dir, use_gnuwin32=False, bin_subdir=False,
 
 def LLVM():
   buildbot.Step('LLVM')
-  Mkdir(LLVM_OUT_DIR)
-  cc_env = BuildEnv(LLVM_OUT_DIR, bin_subdir=True)
+  out_dir = os.path.join(work_dirs.GetBuild(), 'llvm-out')
```
sbc100: How about:

```build_dir = os.path.join(work_dirs.GetBuild(), 'llvm')```

Comment (dschuff) [reply src/build.py]: Currently the default for `GetBuild` is just `src/work` and `src/work/llvm` is the LLVM source directory. So that would be a break from the current configuration. We can talk about doing that but maybe not in this PR.

Comment (sbc100) [reply src/build.py]: Oh i see.  It certainly makes sense to me that if you going distinguish between these different directories that they should not overlap.

File: src/build.py [hunk 2]
```diff
@@ -1693,6 +1706,15 @@ def main():
   if options.torture_filter:
     compile_torture_tests.test_filter = options.torture_filter
 
+  if options.sync_dir:
+    work_dirs.SetSync(options.sync_dir)
+  if options.build_dir:
+    work_dirs.SetBuild(options.build_dir)
+  if options.test_dir:
+    work_dirs.SetBuild(options.test_dir)
```
jgravelle-google: `SetTest` and `SetInstall` ?

.gitignore
```diff
@@ -1,6 +1,9 @@
 *~
 *.pyc
-work/
```
jgravelle-google: Probably want to keep `work/` here if we're not changing the defaults yet?

dschuff: src/work is the correct directory, work/ is currently empty.

dschuff:
OK, this should be reviewable now. the work_dirs abstration is used to remove all the hardcoded build and install directories, but not yet the source directories.

dschuff:
Any more comments on this?

PR #477, opened 2019-03-27
Remove more hardcoded working directory references

dschuff:


src/build.py
```diff
@@ -178,16 +167,34 @@ def CMakeBinDir():
...
+    return GetBuildDir(node_subdir)
+  return GetBuildDir(node_subdir, 'bin')
+
+
+def NodeBin():
+  return Executable(os.path.join(NodeBinDir(), 'node'))
+
+
 # `npm` uses whatever `node` is in `PATH`. To make sure it uses the
 # Node.js version we want, we prepend `NODE_BIN_DIR` to `PATH`.
```
sbc100: Update comment

src/build.py
```diff
@@ -604,49 +611,50 @@ def NoSync(*args):
...
-           WASM_GIT_BASE + 'wabt.git'),
-    Source('binaryen', BINARYEN_SRC_DIR,
-           WASM_GIT_BASE + 'binaryen.git'),
-    Source('musl', MUSL_SRC_DIR,
-           MUSL_GIT_BASE + 'musl.git',
-           checkout=RemoteBranch('wasm-prototype-1')),
-    Source('java', '', '',  # The source and git args are ignored.
-           custom_sync=SyncPrebuiltJava)
-]
+def AllSources():
```
sbc100: Why change this to a function?

src/build.py
dschuff: Done

src/build.py
dschuff: The hardcoded paths above are switched to functions so they can be evaluated after the command-line flags are parsed (since the flags can now change the working directories). Likewise with `AllSources`; since it uses the source directories, it has to be called after flag parsing rather than on startup.

dschuff:
Including for source paths, prebuilt clang, CMake, node, JSVU, Java, and d8.

PR #483, opened 2019-03-29
Remove -DLLVM_COMPILER_CHECKED=ON from libcxxabi build

sbc100:
This is no longer needed after: https://reviews.llvm.org/D58333


dschuff:
Unfortunately the libcxxabi build is still broken because the CMake file sets POSITION_INDEPENDENT_CODE unconditionally and I haven't figured out how to override that from the command line, or if it's even possible.

sbc100:
I'm also looking into the same thing.. any idea why that happens for libcxxabi but not for libcxx?

PR #488, opened 2019-04-09
Use Build dir instead of Sync dir for prebuilts

dschuff:
Prebuilt host tools (clang, CMake) are logically more like sources
than build artifacts, but put them in the build dir anyway because
on LUCI builders, the src dir will be managed by gclient, backed
by git cache, and cleared on each build (whereas the build dir
can be persisted on the builder for incremental compiles)

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -93,7 +93,7 @@ def GetBuildDir(*args):
 
 
 def GetPrebuilt(*args):
-  return GetSrcDir(*args)
+  return GetBuildDir(*args)
 
 
 def GetPrebuiltClang(binary):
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -93,11 +93,15 @@ def GetBuildDir(*args):
...
 
 
 def GetPrebuiltClang(binary):
-  return GetPrebuilt('chromium-clang', 'third_party', 'llvm-build',
+  # For now use the src dir instead of the build dir, because the update script
+  # is managed by gclient.
+  # TODO: copy clang into the build dir so it gets persisted, or figure out how
+  # other bots do it.
+  return GetSrcDir('chromium-clang', 'third_party', 'llvm-build',
                      'Release+Asserts', 'bin', binary)
```
kripken: indentation doesn't look right any more

PR #492, opened 2019-04-17
Switch from jfb's musl port to wasi-sysroot

sbc100:


sbc100:
@dschuff @sunfishcode .. initial CL to switch from jf's musl to wasi-sysroot.

So far compiling and validating seems find module missing "setjmp.h" and "pthread.h" which only effects 6 of out torture tests.

I yet to try running anything..

sbc100:
This should be ready to land now

src/build.py
```diff
@@ -57,7 +57,7 @@
 GIT_MIRROR_BASE = 'https://chromium.googlesource.com/'
 LLVM_MIRROR_BASE = 'https://llvm.googlesource.com/'
 GITHUB_MIRROR_BASE = GIT_MIRROR_BASE + 'external/github.com/'
-WASM_GIT_BASE = GITHUB_MIRROR_BASE + 'WebAssembly/'
+WASM_GIT_BASE = 'https://github.com/WebAssembly/'
```
sunfishcode: If this was the only use of `GITHUB_MIRROR_BASE`, can you remove it entirely?

src/cmake/Modules/Platform/Wasi.cmake
```diff
@@ -1,12 +1,12 @@
...
+# WASI (WebAssembly System Interface) is an experimental WebAssembly standalone
 # toolchain.
 
 # This is arbitrary, AFAIK, for now.
 cmake_minimum_required(VERSION 3.4.0)
 
 set(CMAKE_SYSTEM_VERSION 1)
 set(CMAKE_SYSTEM_PROCESSOR wasm32)
-set(triple wasm32-unknown-unknown-wasm)
+set(triple wasm32-unknown-wasi)
```
sunfishcode: This can be just "wasm32-wasi" now.

src/build.py
```diff
@@ -1137,53 +1140,34 @@ def LibCXXABI():
...
-def Musl():
-  buildbot.Step('musl')
-  build_dir = os.path.join(work_dirs.GetBuild(), 'musl-out')
-  Mkdir(build_dir)
+def Wasi():
+  buildbot.Step('Wasi')
+  build_dir = os.path.join(work_dirs.GetBuild(), 'wasi-sysroot-out')
+  if os.path.isdir(build_dir):
+    Remove(build_dir)
+  cc_env = BuildEnv(build_dir, use_gnuwin32=True)
```
sunfishcode: Can we drop the `use_gnuwin32=True`? We don't have any "gnuwin32" paths in the WASI sysroot.

sbc100: I think this is so that certain tools can be used by the Makefile?  At least they were need for musl. Maybe even needed for make itself? 

I could try changing this as a followup and see how it effects the windows build.

dschuff: This is for tools used by the musl build script (on the host machine). If it breaks you'll see it on the windows bot.

PR #493, opened 2019-04-17
Enable bulk-memory when validating objects

tlively:
My local tests still aren't working, but at least this gets rid of the unknown section errors in the torture tests.


sbc100:
I see.. we are only validating object here.  i think its fine for the objects to be validated with all features enabled.

PR #494, opened 2019-04-17
Fix enabling of bulk-memory during validation

sbc100:
This was attempted in #493 but untested.


src/build.py
```diff
@@ -1422,6 +1427,7 @@ def TestBare():
   for opt in BARE_TEST_OPT_FLAGS:
     CompileLLVMTorture(GetTortureDir('o', opt), opt)
     ValidateLLVMTorture(GetTortureDir('o', opt), 'o', opt)
+  return
```
jgravelle-google: That seems not-ideal?

PR #496, opened 2019-04-24
Switch to llvm monorepo for build trigger

sbc100:


sbc100:
Is there some other central location that I need to update as well?

dschuff:
Not that I know of?

The main effect of this change will be that we'll trigger on a lot more commits (all of the monorepo projects rather than just LLVM). It will make the console view a lot faster/busier. Maybe that's ok? We can try it out.

rryk:
AFAIK, those are the only configs that need to be updated.

PR #500, opened 2019-05-02
Add directory override for V8

dschuff:
This allows V8 to be managed separately by the same gclient that checks out
emscripten-releases and be in a parallel directory (separate from the other
sources).

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1580,6 +1580,9 @@ def ParseArgs():
       '--prebuilt-dir', dest='prebuilt_dir',
       help='Directory for prebuilt output')
   parser.add_argument(
+      '--v8-dir', dest='v8_dir',
+      help='Directory for V8 checkout/build')
+  parser.add_argument(
       '--test-dir', dest='test_dir', help='Directory for test output')
   parser.add_argument(
       '--install-dir', dest='install_dir',
@@ -1700,6 +1703,8 @@ def main():
     work_dirs.SetSync(options.sync_dir)
   if options.build_dir:
     work_dirs.SetBuild(options.build_dir)
+  if options.v8_dir:
+    work_dirs.SetBuild(options.v8_dir)
   if options.test_dir:
     work_dirs.SetTest(options.test_dir)
   if options.install_dir:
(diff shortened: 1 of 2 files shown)


sbc100:
Is that because you want to pin v8?

dschuff:
Yes, the emscripten-releases builder pins V8 via DEPS, and it wants to put v8 in the top-level gclient solution alongside emscripten-releases itself. (https://chromium-review.googlesource.com/c/emscripten-releases/+/1592100/)

sbc100:
Are you going to want to pin all the other things too?

dschuff:
Yes, everything is pinned already (see https://chromium.googlesource.com/emscripten-releases/+/refs/heads/master/DEPS). The recipe just runs build.py with the --no-sync flag and the existing source dir override covers the rest of the sources. V8 is just special because it has DEPS of its own.

sbc100:
Ah, I see.  I still don't see why V8 can't be alongside the rest of the sources?  Does gclient not like having another gclient in a sub-directory?

dschuff:
Basically yes. Specifically in order to build v8 we need to run its DEPS hooks. By default gclient doesnt run hooks recursively (without the change in https://chromium-review.googlesource.com/c/emscripten-releases/+/1592100). V8's hooks work when v8 is at the top level of the gclient-managed directory, but not when a subdir for some reason.

sbc100:
This is kind of sad.   Isn't the v8 we checkoug at the "top" in the sense that it has its own `.gclient` file?

dschuff:
It should be. In principle we'd just want to do in the recipe what we do currently, which is just use a completely different gclient file in a subdirectory, sync it separately, etc. But in recipes, I haven't been able to get the simple approach of "just shell out to gclient" to work; instead it's expected to use the gclient recipe code, which is complex, and I haven't been able to get it to work (another complication is that unlike on the waterfall, we need an unmanaged solution which pins the revision in waterfall's DEPS rather than checking out the latest). In that sense we are more like how Chrome embeds V8, but with the difference that Chrome basically just has its own DEPS hooks which duplicate V8's (which I am trying to avoid). So the current approach has those properties (pinned revision, no duplication of DEPS hooks, works without complex recipe code).

dschuff:
(btw any objections to this PR in particular?)

PR #501, opened 2019-05-03
Run make with --output-sync

sbc100:
This is an attempt fix the issue we are seeing the bots with make
is error'ing out with `make: write error: stdout`


dschuff:
Huh, TIL about that flag.

PR #512, opened 2019-05-17
Prepare for emscripten-releases uploads

kripken:
Allow detecting if we are the emscripten-releases bot, and upload to the proper bucket based on that.

### DIFF
--- a/src/buildbot.py
+++ b/src/buildbot.py
@@ -37,6 +37,15 @@ SCHEDULER = SCHEDULERS[BUILDBOT_SCHEDULER]
 BUILDBOT_REVISION = os.environ.get('BUILDBOT_REVISION', None)
 BUILDBOT_BUILDNUMBER = os.environ.get('BUILDBOT_BUILDNUMBER', None)
 BUILDBOT_BUILDERNAME = os.environ.get('BUILDBOT_BUILDERNAME', None)
+BUILDBOT_MASTERNAME = os.environ.get('BUILDBOT_MASTERNAME', None)
+
+# Possible masters include None (running locally), the waterfall integration
+# bot, or the emscripten-releases bot.
+WATERFALL_BOT = 'client.wasm.llvm'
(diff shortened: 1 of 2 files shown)


src/buildbot.py
```diff
@@ -48,6 +57,12 @@ def BuildNumber():
   return BUILDBOT_BUILDNUMBER
 
 
+def IsEmscriptenReleasesBot():
+  """Return true if running on the emscripten-releases builders,
+     False otherwise."""
+  return BUILDBOT_MASTERNAME is EMSCRIPTEN_RELEASES_BOT
```
dschuff: let's use `==` here for consistency with the rest of this repo.

kripken: Done.

src/buildbot.py
```diff
@@ -37,6 +37,15 @@
...
 BUILDBOT_BUILDNUMBER = os.environ.get('BUILDBOT_BUILDNUMBER', None)
 BUILDBOT_BUILDERNAME = os.environ.get('BUILDBOT_BUILDERNAME', None)
+BUILDBOT_MASTERNAME = os.environ.get('BUILDBOT_MASTERNAME', None)
+
+# Possible masters include None (running locally), the waterfall integration
+# bot, or the emscripten-releases bot.
+WATERFALL_BOT = 'client.wasm.llvm'
+EMSCRIPTEN_RELEASES_BOT = 'emscripten-releases'
+
+assert BUILDBOT_MASTERNAME in [None, WATERFALL_BOT, EMSCRIPTEN_RELEASES_BOT], \
```
dschuff: I realized we'll also need to pass the info about whether the build is a CI build or a try build, and only try to upload from CI builds. I added `BUILDBOT_BUCKET` to the env (which IIUC should be either "ci" or "try").

kripken: Thanks, I'll open a PR.

PR #517, opened 2019-05-20
Don't run wasi on windows

kripken:
It is broken and we're more interested to focus on getting other tests passing on windows.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1373,7 +1373,7 @@ def AllBuilds():
       Build('fastcomp', Fastcomp),
       Build('emscripten', Emscripten),
       # Target libs
-      Build('wasi-sysroot', Wasi),
+      Build('wasi-sysroot', Wasi, os_filter=Filter(exclude=['windows']),
       Build('compiler-rt', CompilerRT),
       Build('libcxx', LibCXX),
       Build('libcxxabi', LibCXXABI),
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -1373,7 +1373,7 @@ def AllBuilds():
       Build('fastcomp', Fastcomp),
       Build('emscripten', Emscripten),
       # Target libs
-      Build('wasi-sysroot', Wasi),
+      Build('wasi-sysroot', Wasi, os_filter=Filter(exclude=['windows'])),
```
dschuff: I think we also have to disable compiler-rt, libcxx and libcxxabi (since those get built into the was sysroot) as well as the 'bare' tests (since those test the wasi components)

sunfishcode:
Could you briefly describe the problems you're seeing, so we can know what it is we need to fix to re-enable this?

kripken:
@sunfishcode - it's probably just some weirdness on the windows bot here, we're not sure. here's an example log:
```
https://logs.chromium.org/logs/wasm/buildbucket/cr-buildbucket.appspot.com/8912917972680810928/+/steps/Wasi/0/stdout
```
So I don't think we know what needs to be fixed here - when we move to the new bots, maybe there won't be anything to fix ;) Or, we'll re-enable there and see how things are.

src/build.py
kripken: @dschuff thanks, I added exclusions for those, PTAL

PR #525, opened 2019-05-22
Enable emscripten testsuite tests on windows

dschuff:


dschuff:
I'm sure this won't work, but running failing tests is better than not running any.

PR #534, opened 2019-05-30
Implement more of the WASI syscalls interface in wasi.js

sbc100:
We now have a lot more filesystem support as well as support
argv and environ.


sbc100:
Oh great!  I was going to say perhaps we could move this implementation somewhere more useful.  I looked for a JS WASI implementation but did find that one.

PR #539, opened 2019-06-12
Don't link libtinfo

kripken:
This may prevent nice colors in the output, but I didn't see that in my local testing (there's no error checking for those flags, though, so not sure I did it right).

Fixes https://github.com/emscripten-core/emsdk/issues/252

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -807,6 +807,9 @@ def LLVM():
       '-DLLVM_ENABLE_ASSERTIONS=ON',
       '-DLLVM_TARGETS_TO_BUILD=X86;WebAssembly',
       '-DLLVM_ENABLE_PROJECTS=lld;clang',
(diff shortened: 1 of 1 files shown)


sbc100:
This seems like fixing one symtom, but the real problem is that we building against out system libraries on our build machine, which means we can end up requiring any libraries from that particular linux release.

At a minimum we should track the minimum requirements of the SDK binaries.   Ideally we'd build against a specific sysroot from a fixes linux version, but that might be overkill.

dschuff:
I do think building in a sysroot (e.g. the Chrome sysroot) does make sense, but it doesn't necessarily solve all the problems. For example at least one of the users' issues was that their version of libtinfo was too new (i.e. it didn't have backwards compatibility like libc does). So statically linking some of the libraries could still help independently of building in a sysroot.

PR #542, opened 2019-06-20
Start to package emscripten libraries in the zips

kripken:
This seems to work locally, but not sure it will work on the bots.

So far just libc, to start with.

See emscripten-core/emscripten#8795 for background discussion


kripken:
The emsdk side is a little nontrivial, yeah. In part due to how caching etc. works there. I have some ideas, I'll start to experiment once this part works ok.

PR #543, opened 2019-06-25
Package generated_struct_info.json and optimizer.2.exe

kripken:
The first experiment of packaging `libc` appears to have worked - I verified emscripten github CI no longer builds libc.a, and tests look good there. Adding two more important files here,

 * generated_struct_info.json is generated in a boostrapping operation that is confusing to debug if the user has a broken setup.
 * optimizer.2.exe takes a while to build and also requires a local compiler environment for native executables.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1015,11 +1015,18 @@ def Emscripten():
     with open(outfile, 'w') as config:
       config.write(text)
 
+  # We don't bundle every single file in the cache, since there are a lot and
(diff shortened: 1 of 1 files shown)


kripken:
Yes, I believe we have a separate dir for each kind - we package the `wasm-obj` and `asmjs` kinds (default, so most commonly used probably).

Waiting a bit before landing this, to see no surprise problems show up on CI.

kripken:
A few more builds passed by, all seem good. Landing...

PR #544, opened 2019-06-25
Build asm.js optimizer in release mode

kripken:
I verified that at least locally for me, without this that executable is built with no optimizations.


dschuff:
wow I wonder how much build-time performance we were leaving on the table over this.

PR #545, opened 2019-06-25
Fix packaging of the asm.js optimizer

kripken:
The asm.js optimizer is built in a special way on the bots, it turns out, and so copying it is special as well (apparently from the comment, building it normally doesn't work... and indeed when I tried to do so it failed).

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1078,12 +1078,6 @@ def Emscripten():
       shutil.copy2(os.path.join(EMSCRIPTEN_CACHE_DIR, cache_subdir, name),
                    os.path.join(packaging_dir, name))
 
-    # The asm.js optimizer is built in a special way above in this function,
-    # and so copying it is special as well.
-    if config_name == 'asm2wasm':
-      shutil.copy2(GetInstallDir('bin', 'optimizer'),
-                   os.path.join(packaging_dir, 'optimizer.2.exe'))
-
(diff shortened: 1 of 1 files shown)


kripken:
After this the waterfall can hopefully get rolled into.

src/build.py
```diff
@@ -1071,10 +1071,18 @@ def WriteEmscriptenConfig(infile, outfile):
       del os.environ['EM_CONFIG']
 
     # Copy the main system libraries so users don't need to themselves.
-    os.makedirs(GetInstallDir('lib', cache_subdir))
-    for cache_lib in cache_libs:
-      shutil.copy2(os.path.join(EMSCRIPTEN_CACHE_DIR, cache_subdir, cache_lib),
-                   GetInstallDir('lib', cache_subdir, cache_lib))
+    packaging_dir = GetInstallDir('lib', cache_subdir)
+    if not os.path.exists(packaging_dir):
```
dschuff: could use os.path.isdir instead of os.path.exists

dschuff: Actually is os.makedirs a no-op for pre-existing directories?

kripken: It's not a no-op - it throws for me locally during testing, but not on the bots where I guess they start fresh.

dschuff: ah ok, then isdir is probably the right thing.

kripken:
Updated to just do the struct info stuff. Will look into a nicer way to do the optimizer, including emsdk changes.

PR #546, opened 2019-07-01
Fix llvm build on older (10.12 and below) OSX versions

sbc100:
See https://github.com/emscripten-core/emscripten/issues/5418

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -815,6 +815,11 @@ def LLVM():
       '-DLLVM_ENABLE_TERMINFO=%d' % (not IsLinux()),
   ])
 
+  if IsMac():
+    # Workaround for https://github.com/kripken/emscripten/issues/5418.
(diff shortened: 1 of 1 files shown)


dschuff:
We could drop support. The OS is 3 years old and according to wikipedia, Apple is dropping support for 10.12 in September of this year. Or we could land this and maybe remove it in September.

sbc100:
I did put that in a comment.   How about we go with this and if it comes up again we can drop.

BTW, how do we pin to a particualar OSX SDK in the OS X build?  IIRC there is a way to set a min ABI version when building on a mac?

dschuff:
IIRC there's a flag -osx-version-min or something like that but I don't think we are using it explicitly. I also don't see it in our mac build output.

sbc100:
I imagine there will be a cmake options that ends up setting this indirectly.

PR #551, opened 2019-07-04
Update tests for indirectbr change

dschuff:


src/test/emwasm_compile_known_gcc_test_failures.txt
```diff
@@ -51,13 +45,10 @@
 
 
 # Computed gotos (but works with bare wasm clang)
-	20041214-1.c O0
-	20071220-1.c
+	20071220-1.c O3
 	20071220-2.c
```
aheejin: Why do we still have computed gotos?

src/test/emwasm_compile_known_gcc_test_failures.txt
```diff
@@ -51,13 +45,10 @@
...
 	990208-1.c # no computed goto, but a blockaddress instruction that we can't select
 
-# Untriaged
-	pr56982.c O3
```
aheejin: This is not a computed goto thing but setjmp fix: #550

src/test/emwasm_compile_known_gcc_test_failures.txt
dschuff: Turns out if you do a blockaddress without an indirectbr the indirectbr lowering pass doesn't lower it away, which makes sense. That still can't really be supported in wasm (e.g. there is no address per se, so no use of the block address other than as a branch target makes sense).

src/test/emwasm_compile_known_gcc_test_failures.txt
```diff
@@ -5,33 +5,30 @@
...
 	20030501-1.c
-	20040302-1.c
 	20040308-1.c
 	20040423-1.c
 	20040520-1.c
 	20041218-2.c
 	20061220-1.c
 	20070919-1.c
-	20071210-1.c
+  20071220-1.c O3
```
aheejin: Oh no I tried to fix the indentation errors by 'resolve conflicts' tool, which didn't show me tabs, and now they are even more unaligned no

PR #552, opened 2019-07-07
Install libclang_shared.so* into wasm-install/lib

aheejin:
After https://github.com/llvm/llvm-project/commit/2e97d2aa1bd313350e76be64299f19172a346bf9,
`CLANG_LINK_CLANG_DYLIB` is `ON` whenever `LLVM_LINK_LLVM_DYLIB` is
`ON`. This requires `libclang_shared.so` to be installed in
`wasm-install/lib`, but `-DLLVM_INSTALL_TOOLCHAIN_ONLY=ON` currently
disables its installation. This adds `libclang_shared.so` to `extra_lib`
so it can be installed. This fixes the current emscripten build step
failure in the waterfall.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -766,7 +766,7 @@ def CopyLLVMTools(build_dir, prefix=''):
                     'llvm-link', 'llvm-mc', 'llvm-nm', 'llvm-objdump',
                     'llvm-readobj', 'opt', 'llvm-dwarfdump'])
   extra_libs = ['libLLVM*.%s' % ext for ext in ['so', 'dylib', 'dll']]
-  extra_libs.append('libclang_shared.*')
+  extra_libs.append('libclang_shared.so*')
   for p in [glob.glob(os.path.join(build_dir, 'bin', b)) for b in
             extra_bins]:
     for e in p:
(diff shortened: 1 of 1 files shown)


aheejin:
Windows will not be affected, because we [disable](https://github.com/WebAssembly/waterfall/blob/67da96c0fd6c3e16c14c48272efec270de038e0c/src/build.py#L798) DYLIB in Windows.

sbc100:
Is this a bug in LLVM_INSTALL_TOOLCHAIN_ONLY?  Surly it should never install a binary without installing its dependencies?

dschuff:
Yeah you may be right. The existing python code also copies libLLVM.so, so maybe it was always broken in that way.

aheejin:
That makes sense. I filed a [bug report](https://bugs.llvm.org/show_bug.cgi?id=42575) for that.

sbc100:
It look like there are other missing too:

https://logs.chromium.org/logs/wasm/buildbucket/cr-buildbucket.appspot.com/8908186013386026608/+/steps/emscripten__emwasm_/0/stdout

Seems to be missing libclang-cpp.so.9svn.

aheejin:
Oh god Oh god https://reviews.llvm.org/D64278 🤦🏻‍♀️

sbc100:
Crap.   I guess thats why its something we need to fix upstream :-/

aheejin:
I agree... I'll track the bug

aheejin:
The upstream bug has been fixed and we don't need to copy them anymore: #556

PR #553, opened 2019-07-08
Install libclang_shared.dylib for mac

aheejin:
Previously #552 failed to install *.dylib for mac. (This tries to install the
nonexistent *.dll too just in case we enable DYLIB in Windows too, but if the
file doesn't exist, it's ok)


aheejin:
tbh I thought mac uses *.so too...

tlively:
Nice, I had forgotten about mac as well.

PR #556, opened 2019-07-12
Don't copy extra libraries anymore

aheejin:
The bug that necessary shared libraries were not copied when
`LLVM_INSTALL_TOOLCHAIN_ONLY=ON` and `DYLIB=ON` has been fixed upstream:
https://github.com/llvm/llvm-project/commit/a196469e67ce578df4fc9f348cc5b7221f12b239
https://github.com/llvm/llvm-project/commit/1d062dae158c5ba7d1f27010d5e4178e10820ee2

So we don't need to copy them anymore.


dschuff:
Turns out this is only *mostly* fixed upstream: https://chromium-review.googlesource.com/c/emscripten-releases/+/1700366

PR #558, opened 2019-07-17
Remove em* wrapper scripts. Fixes #557

### LINKED ISSUE #557: The mystery of the /b/s/w/ir/k/i paths on emsdk users' machines, resolved
We've gotten bug reports from users with errors like
```
shared:ERROR: llc executable not found at `/b/s/w/ir/k/install/fastcomp/bin/llc`
```
And it was a complete mystery where such a path comes from. Today I saw that very error on my machine! The culprit is not in emscripten, nor in the emsdk, but here in the waterfall, specifically

https://github.com/WebAssembly/waterfall/blob/master/src/emcc_wrapper.sh

That file ends up in the `bin/` dir, so that after the user installs the emsdk, that file ends up copied into `./upstream/bin/emcc`. (Confusingly, the file has no relation to the `emcc` file in the emscripten repo.) It contains this:
```
export EM_CONFIG="${WASM_ROOT}/emscripten_config_fastcomp"
```
And that config file indeed contains
```
WASM_INSTALL = '/b/s/w/ir/k/install'
```
As those configs are hardcoded for our build machines...

I'd like to not ship this file at all to users (nor those config files), so that `upstream/bin/emcc` will not exist. We don't mention it in our docs. I used it by mistake, I guess like users have, just by searching for "emcc" under the emsdk dir and that being the first result; or, by adding `upstream/bin` to the path, they get that file when they try to run `emcc` (the emsdk does not add it to the path itself).

Does that make sense or have I missed something in this file's use?

cc @sbc100 who wrote that file.

kripken:
I wasn't sure how to check if these are actually used somewhere. If neither of you are sure they aren't used, I can do a full local test (running a quick test I didn't see them used).


kripken:
Checking with @dschuff this sgt as well. Landing, and hopefully no surprises when this tries to roll...

PR #560, opened 2019-07-19
Restore printing of archive contents before upload

dschuff:
This should work again with the new bot infrastructure.
Also remove redundant check in subroutine, which makes things
easier to test locally.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -293,8 +293,6 @@ def CopyLibraryToSysroot(library):
 def Archive(directory, print_content=False):
   """Create an archive file from directory."""
   # Use the format "native" to the platform
-  if not buildbot.IsBot():
-    return
   if IsWindows():
     return Zip(directory, print_content)
   return Tar(directory, print_content)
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -293,8 +293,6 @@ def CopyLibraryToSysroot(library):
 def Archive(directory, print_content=False):
   """Create an archive file from directory."""
   # Use the format "native" to the platform
-  if not buildbot.IsBot():
-    return
```
sbc100: Did you mean to remove this?

src/build.py
```diff
@@ -1208,7 +1206,7 @@ def ArchiveBinaries():
   # TODO(sergiyb): Restore printing list of binaries on Linux and Mac once it
   # works. See https://crbug.com/916775 and https://crbug.com/940663
```
sbc100: Remove the TODO too?

src/build.py
dschuff: Yes, see the "Also" part of the comment.

src/build.py
dschuff: Done.

PR #561, opened 2019-07-19
 Further reduce the extra binaries and libs in the uploaded packages

dschuff:
Remove a few extra binaries and libclang, and use the Executable()
utility function so that things work on Windows too.


dschuff:
Funnily enough, making this logic actually work on windows reduces the package size by over 250M. But there's still a lot more we can do.

dschuff:
@kripken just in case things go awry for emsdk users next week, this would be the PR to revert.

PR #568, opened 2019-08-14
Explain __builtin_return_address situation

quantum5:


quantum5:
The offset converter is used for stack trace purposes in sanitizers, but that is beyond simple `__builtin_return_address`.

PR #571, opened 2019-08-29
Set CMAKE_OSX_DEPLOYMENT_TARGET on mac builds

sbc100:


sbc100:
I don't really know how to test this...

dschuff:
I'll try it locally. You can tell with `otool` what the target is.

dschuff:
```
Load command 9
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.12
      sdk 10.13
```

src/build.py
```diff
@@ -814,11 +817,6 @@ def LLVM():
       '-DLLVM_ENABLE_TERMINFO=%d' % (not IsLinux()),
   ])
 
-  if IsMac():
```
dschuff: I guess we also need to remove the one that you added to fastcomp too?

sbc100:
Presumably this also removes the futimes dependency?

dschuff:
I would assume so. I'd guess that the headers you get when you set 10.12 as the deployment target should match what the patform actually has.

PR #572, opened 2019-09-25
Update CMake to 3.15.3 to get C++17 support

tlively:


src/build.py
```diff
@@ -177,7 +177,7 @@ def CMakeArch():
   return 'x64' if IsWindows() else 'x86_64'
 
 
-PREBUILT_CMAKE_VERSION = '3.7.2'
+PREBUILT_CMAKE_VERSION = '3.15.3'
```
sbc100: Can you set this to 3.7.2-1 since that is the version in debian/stretch (oldstable):  https://packages.debian.org/stretch/cmake?

sbc100: Also, make add a comment above saying why 3.7.2 was chosen (so that we ensure compatibility with people on older distros and that we chose debian/oldstable as the cut off point?)

sbc100: Sorry ignore me... we are on 3.7.2 already.   3.8 didn't make it into a debian/stable release until https://packages.debian.org/buster/cmake which only just came out..

So I think maybe we should change binaryen to not require 3.8.. sorry.

tlively:
Closing in favor of https://github.com/WebAssembly/binaryen/pull/2358

PR #573, opened 2019-10-17
Run a limited set of default builds/tests when invoking build.py without filters

dschuff:
This reduces the number of tests run on the wasm-stat.us waterfall, removing
fastcomp torture tests (which are not worth keeping) and all emscripten
test-suite tests (which we run more comprehensively elsewhere).


dschuff:
This will also allow adding a build step for LLVM test suite which is off by default

PR #575, opened 2019-10-24
Remove natives_blob.bin from v8 build

sbc100:
This was removed completely in:
  https://chromium-review.googlesource.com/c/v8/v8/+/1844771


sbc100:
I guess this isn't breaking on the emscripten-releases waterfall because we don't do clean builds ever?

dschuff:
Yeah, that's probably bad.
OTOH the emscripten-releases waterfall doesn't actually do anything with the v8 it builds right now, either; all the testing is with node.

PR #578, opened 2019-10-31
Split LLVM Regression tests into a separate step from the LLVM build

dschuff:
This will allow splitting into separate build steps on recipes-based bots.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -838,7 +838,17 @@ def LLVM():
         shutil.copy2(Executable(os.path.join(install_bin, target)),
                      Executable(link))
 
+
+
+def TestLLVMRegression():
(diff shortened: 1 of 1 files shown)


sbc100:
Maybe split out v8 tests and then we can remove the "--no-tool-test" option too?

dschuff:
Or we could just leave the v8 tests out. I don't think they are particularly valuable, they've basically never failed on our waterfals.

sbc100:
Sure.  Either way we can remove the `--no-tool-test` option (do we run wabt tests?)

dschuff:
yeah, we don't run wabt tests either

dschuff:
I'll land this and do that in a followup CL.

PR #579, opened 2019-11-01
Split Emscripten build step into upstream and fastcomp variants

dschuff:
More specifically, instead of building libs for both variants, the step
now only builds libs for the specfied variant. This allows the
llvm-test-suite bot to skip building fastcomp altogether.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -993,6 +993,7 @@ def Fastcomp():
 def BuildEmscriptenOptimizer():
   # Remove cached library builds (e.g. libc, libc++) to force them to be
   # rebuilt in the step below.
+  buildbot.Step('emscripten (optimizer)')
   Remove(EMSCRIPTEN_CACHE_DIR)
   src_dir = GetSrcDir('emscripten')
   em_install_dir = GetInstallDir('emscripten')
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -990,34 +990,40 @@ def Fastcomp():
...
+    cc_env = BuildEnv(optimizer_out_dir)
+    command = CMakeCommandNative([
+        os.path.join(src_dir, 'tools', 'optimizer'),
+        '-DCMAKE_BUILD_TYPE=Release'
+    ])
+    proc.check_call(command, cwd=optimizer_out_dir, env=cc_env)
+    proc.check_call(['ninja'] + host_toolchains.NinjaJobs(),
+                    cwd=optimizer_out_dir, env=cc_env)
+    CopyBinaryToArchive(Executable(
+        os.path.join(optimizer_out_dir, 'optimizer')))
```
sbc100: Create a sub-routine called `BuildEmscriptenOptimizer`?

src/build.py
```diff
@@ -1035,61 +1041,65 @@ def WriteEmscriptenConfig(infile, outfile):
...
+  # rewrites the sanity file. Then, Mac OS X and Windows' timestamps for
+  # mtime only have 1~2 seconds resolution, which makes it possible that
+  # other emcc processes launched later by the initial emcc process again
+  # think the cache is invalid and delete the cache again, which would then
+  # contain half-built libraries, if less than 1 seconds has elapsed after
+  # the first deletion. So here we make sure that enough time has elapsed
+  # between configuration file writing and sanity checking, so that it would
+  # temporarily work on Mac and Windows.
+  if not IsLinux():
+    time.sleep(3)
```
sbc100: This might not be needed anymore since #9347?  Separate PR though I guess

PR #584, opened 2019-11-06
Add Git executable to CMake command line on Windows

dschuff:
Git's usual FindPackage logic doesn't find it on LUCI (and git's now required
for wabt)

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -941,10 +941,6 @@ def Wabt():
   ]), cwd=out_dir, env=cc_env)
 
   proc.check_call(['ninja', '-v'], cwd=out_dir, env=cc_env)
-  # TODO(sbc): git submodules are not yet fetched so we can't yet endable
-  # wabt tests.
-  # if options.run_tool_tests:
-  #   proc.check_call(['ninja', 'run-tests'], cwd=out_dir, env=cc_env)
   proc.check_call(['ninja', 'install'], cwd=out_dir, env=cc_env)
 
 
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -936,7 +940,7 @@ def Wabt():
       '-DBUILD_TESTS=OFF'
   ]), cwd=out_dir, env=cc_env)
 
-  proc.check_call(['ninja'], cwd=out_dir, env=cc_env)
+  proc.check_call(['ninja', '-v'], cwd=out_dir, env=cc_env)
```
sbc100: Unrelated?

src/build.py
```diff
@@ -731,9 +731,13 @@ def CMakeCommandBase():
...
   command.append('-DCMAKE_EXPORT_COMPILE_COMMANDS=ON')
   command.append('-DCMAKE_BUILD_TYPE=Release')
-  if IsMac:
+  if IsMac():
     # Target macOS Seirra (10.12)
     command.append('-DCMAKE_OSX_DEPLOYMENT_TARGET=10.12')
+  elif IsWindows():
+    # CMake's usual logic fails to find LUCI's git on Windows
+    git_exe = proc.check_output(['where', 'git']).strip()
+    command.append('-DGIT_EXECUTABLE=%s' % git_exe)
```
sbc100: Oh come on.  Really camke?   How can you not do this.

Also, do we not already a have a find_in_path type function in this project?

dschuff: Wabt uses FIND_PACKAGE (https://github.com/WebAssembly/wabt/blob/master/CMakeLists.txt#L70) but it doesn't seem to work. To be fair, LUCI's setup is weird, especially on Windows.

dschuff: Oh you're right though, we have `proc.which`. I'll try that.

PR #585, opened 2019-11-06
Update to use CMake 3.15 on the waterfall

dschuff:
CMake 3.15 has much nicer support for clang-cl on Windows.


dschuff:
Yeah, I don't plan to change the version we use on Travis/CircleCI. Just for emscripten-releases.

But also... come on, 3.5.1?
FWIW we may have to switch to 3.15 regardless: http://lists.llvm.org/pipermail/llvm-dev/2019-October/136295.html

PR #587, opened 2019-11-07
Compile the host binaries with Clang on Windows

dschuff:
We still use the MSVC SDK from V8, but now we use V8's (separately-DEPSed)
clang on Windows (the same one that we use on the other platforms).

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -41,6 +41,7 @@ import proc
 import testing
 import work_dirs
 
+
 SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
 JSVU_OUT_DIR = os.path.expanduser(os.path.join('~', '.jsvu'))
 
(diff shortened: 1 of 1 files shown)


dschuff:
To land once the current in-flight PRs get rolled, and we tag a release.

src/build.py
```diff
@@ -805,6 +809,7 @@ def LLVM():
   build_dylib = 'ON' if not IsWindows() else 'OFF'
   command = CMakeCommandNative([
       GetLLVMSrcDir('llvm'),
+      '-DCMAKE_CXX_FLAGS=-Wno-nonportable-include-path',
```
sbc100: Why this?

dschuff: There are a few places where the filename on disk differs in case from the filename in the #include statement, which produces a warning. There aren't a lot but they are included indirectly in basically every file, and it results in a lot of log spam.

dschuff:
All the prerequisites seem to be rolled and green, let's give this a shot.

src/build.py
dschuff: Actually I saw this in Binaryen too and looked again. It looks like in that case the name actually does match (e.g. both the [filename](https://github.com/WebAssembly/binaryen/blob/master/src/cfg/Relooper.h) in the repo and the include start with a cap; but the [warning](https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket.appspot.com/8897430022390414848/+/steps/Build_Binaryen/0/stdout) complains that the path on disk has no cap. So maybe there's something about how the repo is cloned or checked out.

sbc100: Was the file renamed at some point?   Perhaps a clobber build?

PR #588, opened 2019-11-07
Unforce the host clang toolchain for wabt

dschuff:
Chromuim's wabt mirror is stuck and out of date, so we can't roll in the
change that makes the wabt build compatible with windows clang.
So for now, fall back to MSVC


dschuff:
TBR @kripken to unstick the waterfall roll

PR #589, opened 2019-11-08
Build and run the LLVM test suite

dschuff:
For now, just the `SingleSource` tests (approximately 300 tests).
There are some current failures, which are listed in the expected_fails file, using the same format as the torture test expectations file.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -245,6 +245,9 @@ LLD_KNOWN_TORTURE_FAILURES = [os.path.join(SCRIPT_DIR, 'test',
 LLVM_TORTURE_EXCLUSIONS = [os.path.join(SCRIPT_DIR, 'test',
                                         'llvm_torture_exclusions')]
 
+RUN_LLVM_TESTSUITE_FAILURES = [os.path.join(SCRIPT_DIR, 'test',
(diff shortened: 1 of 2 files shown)


src/build.py
```diff
@@ -1603,6 +1606,61 @@ def TestEmtestAsm2Wasm():
...
 
+def TestLLVMTestSuite():
+  outdir = GetBuildDir('llvmtest-out')
+  # The compiler changes on every run, so incremental builds don't make sense.
+  Remove(outdir)
+  Mkdir(outdir)
+  # The C++ tests explicitly link libstdc++ for some reason, but we use libc++
+  # and it's unnecessary to link it anyway. So create an empty libstdc++.a
+  proc.check_call([GetInstallDir('bin', 'llvm-ar'), 'rc', 'libstdc++.a'],
+                  cwd=outdir)
```
sbc100: Hmm.. this seems strange/bad.  I wonder how this runs in other libc++ platforms?

dschuff: Not sure. Probably there are already other compatibility libs on the system. or maybe they really want to link against libstdc++ specifically; but that would be weird given that they run this on macs. so :man_shrugging:

PR #590, opened 2019-11-08
Support Python3 in waterfall build scripts

dschuff:


src/parallel_runner.py
```diff
@@ -14,8 +14,12 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+from __future__ import print_function
 import multiprocessing
-import Queue
+try:
+  import queue
+except ImportError:
+  import Queue as queue
```
sbc100: Comment here ?

src/host_toolchains.py
```diff
@@ -66,7 +67,7 @@ def GetVSEnv(dir):
         continue
       var, val = e.split('=', 1)
       env[var] = val
-      print('ENV: %s = %s' % (var, val))
+      print(('ENV: %s = %s' % (var, val)))
```
sbc100: Overzealous search and replace?

src/compile_torture_tests.py
```diff
@@ -57,7 +57,7 @@ def find_runnable_tests(directory, pattern):
     for filename in files:
       if fnmatch.fnmatch(filename, pattern):
         fullname = os.path.join(root, filename)
-        with open(fullname, 'r') as f:
+        with open(fullname, 'r', encoding='ISO8859-1') as f:
```
sbc100: wat?

src/build.py
```diff
@@ -549,21 +552,26 @@ def SyncArchive(out_dir, name, url):
...
   work_dir = os.path.dirname(out_dir)
 
   try:
-    f = urllib2.urlopen(url)
-    print 'URL: %s' % f.geturl()
-    print 'Info: %s' % f.info()
+    if sys.version_info.major == 2:
+      from urllib2 import urlopen, URLError
+    else:
+      from urllib.request import urlopen, URLError
```
sbc100: Why if/else here but  try/except for `queue`

Maybe the if/else is better because it self documenting?

src/build.py
```diff
@@ -125,7 +125,7 @@ def IsWindows():
 
 
 def IsLinux():
-  return sys.platform == 'linux2'
+  return sys.platform == 'linux2' or sys.platform == 'linux'
```
sbc100: `.startswith('linux')`?

src/compile_torture_tests.py
dschuff: Apparently it makes a difference. By default it tries UTF-8 encoding and fails to read several of the tests (the offending characters appear to be in comments)

sbc100: Holy crap.   Can you add a command about this?

src/parallel_runner.py
dschuff: Replaced with explicit version check

src/build.py
dschuff: Yeah I switched the other one for that reason.

PR #593, opened 2019-11-13
Add .vpython3 file

dschuff:
It goes along side .vpython file to isolate the python environment on
Chromium builders

### DIFF
--- a/.vpython3
+++ b/.vpython3
@@ -1,0 +1,24 @@
+# This is a vpython "spec" file.
+#
+# It describes patterns for python wheel dependencies of the python scripts in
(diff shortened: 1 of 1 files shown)


sbc100:
But are we running `vpython` anywhere?   

I'm not a big fan for more chromium-specific stuff here but I guess we are stuck with it for now.

dschuff:
vpython is used by default by the recipe runner. We already have a `.vpython` for its python2 installation, so this really isn't a change. Specifying a `.vpython3` file in the recipe script is the only way I know of to make it use python3. For every other purpose, everything just works if you kick the whole process off by running python3 instead of python2.

PR #599, opened 2019-12-10
Use Python3 in vpython file

dschuff:


dschuff:
cc @tlively @sbc100 going to see if this fixes the mac infra issues

PR #600, opened 2019-12-10
Run v8 build using vpython

dschuff:
This runs v8's python scripts (which need python2) in a separate environment (controlled by v8's `.vpython` file) from the calling environment (which is python3).

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -896,7 +896,7 @@ def V8():
   buildbot.Step('V8')
   src_dir = work_dirs.GetV8()
   out_dir = os.path.join(src_dir, 'out.gn', 'x64.release')
-  proc.check_call([os.path.join(src_dir, 'tools', 'dev', 'v8gen.py'),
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -896,7 +896,8 @@ def V8():
   buildbot.Step('V8')
   src_dir = work_dirs.GetV8()
   out_dir = os.path.join(src_dir, 'out.gn', 'x64.release')
-  proc.check_call([os.path.join(src_dir, 'tools', 'dev', 'v8gen.py'),
+  proc.check_call(['vpython',
```
sbc100: Seems sad to depend on depoy_tools here, but I guess we already implicitly do?

dschuff:
Yes, the v8 build already depends on depot_tools. This (and the follow-on CL) only use vpython when building v8 and when using the win toolchain that comes with depot_tools, so I don't think it's adding any dependencies that aren't already there.

PR #601, opened 2019-12-10
Make vpython-invocation code work on windows

dschuff:
Add .bat if necessary and also update the windows-specific logic
in host_toolchains.py

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -896,7 +896,7 @@ def V8():
   buildbot.Step('V8')
   src_dir = work_dirs.GetV8()
   out_dir = os.path.join(src_dir, 'out.gn', 'x64.release')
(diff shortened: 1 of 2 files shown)


src/build.py
```diff
@@ -896,7 +896,8 @@ def V8():
   buildbot.Step('V8')
   src_dir = work_dirs.GetV8()
   out_dir = os.path.join(src_dir, 'out.gn', 'x64.release')
-  proc.check_call(['vpython',
+  vpython = 'vpython' + '.bat' if IsWindows() else ''
```
tlively: Can we add parens here to make the order of operations clearer?

dschuff: Done

PR #605, opened 2019-12-11
Use vpython as gn's script interpreter

dschuff:
Python scripts executed by gn don't work under python3, so use vpython
to isolate them from build.py's environment. This requires invoking
gn directly rather than using V8's gn scripts.


tlively:
What a pain! Do we know when gn and V8 will move to Python 3?

PR #607, opened 2019-12-12
Use the right number of ninja jobs for goma when building fastcomp

dschuff:
Also remove the duplication of the goma-finding logic in V8,
and simplify CMake command generation now that we don't have
special-cases overriding host clang per-callsite.

### DIFF
--- a/src/host_toolchains.py
+++ b/src/host_toolchains.py
@@ -140,7 +140,7 @@ def CMakeLauncherFlags():
 
 
 def NinjaJobs():
-  if UsingGoma():
+  if UsingGoma() and force_host_clang:
     return ['-j', '50']
   return []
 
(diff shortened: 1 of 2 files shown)


src/host_toolchains.py
```diff
@@ -140,7 +140,7 @@ def CMakeLauncherFlags():
 
 
 def NinjaJobs():
-  if UsingGoma():
+  if UsingGoma() and force_host_clang:
```
tlively: I'm definitely missing context here. It seems to me that `UsingGoma()` and `force_host_clang` should be contradictory.

dschuff: It's actually the other way around. The "forced" clang is the one that comes from the chrome infra, so it's the one that's guaranteed to be in goma. (The "unforced" clang is just whatever the default local compiler is).

PR #608, opened 2019-12-13
Force Python UTF-8 mode when running tests

dschuff:
This overrides the default locale to always use UTF-8 for decoding text
files (on Windows the default is mbcs/cp1252). This is because our test
files are all encoded with UTF-8.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1858,6 +1858,9 @@ def run(sync_filter, build_filter, test_filter):
     Summary()
     return 1
 
+  # Override the default locale to use UTF-8 encoding for all files and stdio
+  # streams (see PEP540), since oure test files are encoded with UTF-8.
+  os.environ['PYTHONUTF8'] = '1'
   for t in test_filter.Apply(ALL_TESTS):
     t.Test()
 


dschuff:
This is a (possibly short-term) alternative to  #10027 which should fix the tests until we figure out the right way to handle encodings with py3.

src/build.py
```diff
@@ -1858,6 +1858,9 @@ def run(sync_filter, build_filter, test_filter):
     Summary()
     return 1
 
+  # Override the default locale to use UTF-8 encoding for all files and stdio
+  # streams (see PEP540), since oure test files are encoded with UTF-8.
+  os.environ['PYTHONUTF8'] = '1'
```
sbc100: Would it make more sense to do this in emscripten itself (i.e. in runner.py)?   Or would that be too late?

dschuff: As far as I can tell there's no supported way to do it once the process has started up. Doing it in runner.py would work for subprocesses but not the parent process. (the test runner does open source files and output e.g. in `run_generated_code` and `do_run_from_file`).  I think there are some situations where there' no supbrocess creation in the test runner.

sbc100: Right.. and are most of the places we are about in the test process, or in the emcc subprocesses?

dschuff:
I'm going to land this to unblock the tests, we can work on a more permanent solution.

PR #612, opened 2020-01-15
Update to llvm 11

sbc100:


sbc100:
Aj!, maybe its not needed by emscripten-releases.. but its used in the libcxx build on the other waterfall:
https://logs.chromium.org/logs/wasm/buildbucket/cr-buildbucket.appspot.com/8891156454920845984/+/steps/libcxx/0/stdout

PR #613, opened 2020-01-17
Git clone and check including submodules

sbc100:
This fixes the wabt build that now depends on submodules


sbc100:
No the main build now depend on them.   (To build libwasm.so C-API impl).

I guess the chrome infra sync is including them already?

PR #615, opened 2020-01-17
Second attempt to get submodules working

sbc100:


sbc100:
Actually I think the reason this isn't breaking on the emscripten-releases tree is that we don't build wabt there.

dschuff:
Except that we do: https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket.appspot.com/8891042438656248064/+/steps/Build_Wabt/0/stdout

sbc100:
From the docs for git checkout:

```
       --recurse-submodules, --no-recurse-submodules
           Using --recurse-submodules will update the content and current branch of all initialized submodules in addition
           to the superproject. If local modifications in a submodule would be overwritten, the checkout will fail unless
           -f is used. If nothing (or --no-recurse-submodules) is used, the work trees of submodules will not be updated.
```

The problem is that `all initialized submodules`.  If somebody adds a new submodule and you checkout the branch (as is the case with the current problem) that new submodule is not initialized and therefore will not be updated.

PR #617, opened 2020-01-22
yapf-format all the things

sbc100:
Switch to 4 space indentation.  We already switched in wabt and
binaryen.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
-ignore = W504,E114,E111
+ignore = W504
 exclude = ./debian,./src/work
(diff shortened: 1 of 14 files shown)


sbc100:
I did do a few post-yapf cleanup for cases where I think yapf was confused.. or at least the output conflicted with pep8.

Maybe look at this diff with whitespace changes disabled

dschuff:
I'm not super happy about all the places where it wants to blow lists/params up onto so many more lines, but in general I'm in favor of auto-formatting if it's easy to keep using it continuously. What post-yapf cleanup was required, and will it keep having to be done repeatedly (or is there a feature for handling just diffs like clang-format?)

PR #618, opened 2020-01-27
Run `npm install` before running emscripten tests

sbc100:


dschuff:
Looks like this isn't working: https://ci.chromium.org/p/emscripten-releases/builders/try/linux/b8890072294421298416
although that's odd because we put the node Bin dir in the path: https://github.com/WebAssembly/waterfall/blob/master/src/build.py#L170

src/build.py
```diff
@@ -1615,6 +1615,8 @@ def ExecuteEmscriptenTestSuite(name, tests, config, outdir, warn_only=False):
   buildbot.Step('Execute emscripten testsuite (%s)' % name)
   Mkdir(outdir)
   try:
+    proc.check_call([os.path.join(NodeBinDir(), 'npm'), 'install'],
```
dschuff: also. does this need to be wrapped in `Executable()`?

sbc100: Yup .. I think so.

sbc100:
I don't get it either..

PR #619, opened 2020-01-27
Attempt to debug node path issues

sbc100:
Also, attempt to debug "can't find node" issue in emscripten-releases
builder:
https://ci.chromium.org/p/emscripten-releases/builders/try/linux/b8890072294421298416

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -945,11 +945,11 @@ def Jsvu():
     os_id = 'linux64'
     js_engines = 'javascriptcore'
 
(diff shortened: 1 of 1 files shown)


dschuff:
Actually looking on my windows machine, it looks like 'npm' is a batch file and not an exe. So maybe `Executable` is not what we want (maybe `shell=True`?) I'll try it on my windows machine and see.

sbc100:
Ok reverted that part.

PR #620, opened 2020-01-28
Attempt to fix execution of `npm` on windows builder

sbc100:


src/proc.py
```diff
@@ -56,7 +56,7 @@ def MungeExe(cmd, cwd):
   if exe.endswith('.py'):
     script = Which(exe, cwd, is_executable=False)
     return [sys.executable, script] + cmd[1:]
-  if exe == 'git' or exe == 'gclient':
+  if exe in ('git', 'npm', 'gclient'):
```
dschuff: Instead of special-casing 'npm' here, we could just do `Executable('npm', '.cmd')` above in the check_call invocation. I guess this is OK since npm is used twice (although not on Windows, but still)

PR #623, opened 2020-01-31
Download MSVC 2010 redist DLL for closure on Windows

dschuff:


sbc100:
maybe not needed if we land this? : https://github.com/emscripten-core/emscripten/pull/10328

sbc100:
We can't expect all our users to do this right?

dschuff:
Right; although the bit we are doing now with `npm ci` also does not affect any users. Currently we are either expecting them to manage closure completely on their own (unless emsdk handles closure?)

dschuff:
On the flipside if we get this working we can stop using Java on our bots entirely, at the risk of diverging from the version of closure most of our users use (assuming they use Java)

sbc100:
emsdk handles the `npm ci` so we would need to replicate this fix there too if we go with it.

I prefer sticking with java for the windows users until upstream fixes the windows package.

sbc100:
I landed the emscripten fix so closing this one

PR #624, opened 2020-02-05
Limit llvm build for WebAssembly only

sbc100:
This will help speed up the build and test of llvm and should
also avoid confusion when users see x86 support listed (and even
the default) in the underlying compiler.


dschuff:
The problem with this is that when I last tested it locally, a lot of lit tests failed. The tests don't seem to be well-prepared for a cross-only compiler where the native target isn't supported. Have you tried that locally? If it works I'm happy to do this though.

sbc100:
No.. I'm not tested.  Will do.

dschuff:

PR #625, opened 2020-02-12
Update expectaions after llvm change

sbc100:
This llvm change make these two tests start failing in O2 and O3
where-as before they were only failing in O0:

 https://reviews.llvm.org/D61314


dschuff:
LGTM

PR #627, opened 2020-02-19
Update emscripten cache directory

sbc100:
The directory within the cache was renamed in emscripten to just "wasm".


sbc100:
Actually this will need a combined roll :(

PR #628, opened 2020-02-20
Remove now unnecessary hack

sbc100:
I believe this issue was fixed in:
 https://github.com/emscripten-core/emscripten/pull/9347

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1096,27 +1096,6 @@ def Emscripten(variant):
   src_config = os.path.join(SCRIPT_DIR, os.path.basename(config))
   WriteEmscriptenConfig(src_config, config)
 
-  # FIXME HACK emcc does sanity check by comparing mtime of the sanity file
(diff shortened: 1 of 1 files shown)


sbc100:
I think I see what you mean. 

In order for that to happen emcc would have to run (in order to generate the sanity file), and then some outside actor would need to modify the settings and then run-run emcc right away such that the sanity and the settings would have the same timestamp?

I *think* that is a potential issue, although it doesn't seem to happy in our tests, and we do quite a lot of testing around this.

I have a hard time imagining any real user hitting this..  can you imagine a real situation that might trigger this?

tlively:
Yeah, it seems possible to do, but very hard to do accidentally. What's the worst that will happen if you run with an invalidated cache?

sbc100:
I think the worst side effect is basically link failure because you pre-built libraries don't match your headers.

PR #629, opened 2020-02-20
Remove support for python2

sbc100:
Seems like the time has come

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
       - ninja-build
       - devscripts
 python:
-  - "2.7"
+  - "3.6"
 install:
   - pip install flake8==3.7.8
 script:
--- a/__init__.py
+++ b/__init__.py
@@ -1,5 +1,3 @@
-#! /usr/bin/env python
-
 #   Copyright 2015 WebAssembly Community Group participants
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,5 +1,3 @@
-#! /usr/bin/env python
-
 #   Copyright 2015 WebAssembly Community Group participants
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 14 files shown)


dschuff:
We haven't actually pulled the trigger on Py2 in emscripten yet, correct? Do we want to wait on this until we do that?

sbc100:
My plan was to start removing python2 and using py3-only features one project at time.   

I did wabt already: https://github.com/WebAssembly/wabt/pull/1321

Next up is binaryen.   Not that for wabt and binaryn the dependency is only on those that build the projects, not those that run them.

I imagine emscripten and emsdk will probably take longer since we have more end users running the python code directly.

Maybe I'm misunderstanding.  Why would we want to block converting this project to python3 on converting emscripten?

dschuff:
If we need to run Python2 on our CI in order to test emscripten in python2, then it might be convenient to have the waterfall support python2. Otherwise we'd have to run the waterfall under python3 and then invoke python2 explicitly when running emscripten (and make sure that the environment is correct e.g. using vpython if applicable). Not impossible, just maybe more trouble than just keeping python2 in the waterfall.

dschuff:
(I guess this is less of a problem on CircleCI than it is on chromium CI? So maybe we just only use py3 on Chromium CI?)

sbc100:
I think its fine for the test running (the waterfall) to be python3 while the system under test (emscripten) to be using python2.

If we want to be able to test emcc with both python3 and python3 then we have to pick just one to run the waterfall code under, right?

dschuff:
I think currently emcc will just run under whatever we run the waterfall under, but I'm not sure if that's just because chromium CI only set up the environment for one of them? Currently the waterfall at least runs the emscripten test runner explicitly under sys.executable.

sbc100:
Yes, it goes look like it used `sys.executable` to run `tests/runner.py` which in turn will use `sys.executable` to run `emcc.py`.

However if we want to support both we probably want to control that explicitly from build.py..  I think `sys.executable` would not be relevant in that world anyway, once we make it explicit.  

I still this this PR is fine, if we want to being back python2 testing we would just modify the callsite where `runner.py` is run.

dschuff:
Yeah, agreed.

PR #631, opened 2020-03-27
Use emscripten's install.py to copy emscripten

sbc100:


sbc100:
Oh darn, I forgot about that.  I was hoping just the linux bot would be able to install emscripten and that could be installed on mac and windows... I forgot we do monolithic builds now :(

dschuff:
The system we use for the PNaCl toolchain can do that, but I'm not sure I want to import that here, it's 10x more complex.

sbc100:
If we switch back to a component system in emsdk we could do this... but that is a bunch of work.

sbc100:
Any chance we can start depending on WSL on windows any time soon?

dschuff:
Do they make it easy to build windows binaries that don't need WSL to run?
not sure about chrome infra though, I doubt they have it now.

dschuff:
Also it wouldn't really help for the case where we actually want to run tests on windows, in an environment like our users have, right? We still need to copy the new emscripten files somehwere to use them for tests.

PR #632, opened 2020-03-27
Update to travis testing to xenial

sbc100:
It seems the gcc version on trusty is no longer capable of building
wabt.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: false
-dist: trusty
+dist: xenial
 language: python
 addons:
   apt:


sbc100:
Not sure why travis jobs are not showing up here.. are we just waiting in travis to start?

binji:
> Not sure why travis jobs are not showing up here.. are we just waiting in travis to start?

I've noticed the same thing with travis on wabt recently too.

sbc100:
I think we should just skip that an jump to github actions: https://github.com/WebAssembly/waterfall/pull/633

PR #634, opened 2020-04-01
Build Linux binaries in V8's sysroot

dschuff:
This uses a hermetic build environment that will have broader
compatibility than whatever happens to be on the buildbots.

Also disable libxml2 in the LLVM build; for some reason CMake has trouble finding
it in the sysroot, but it's only needed for generating Windows manifests which we 
don't need.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,4 +33,4 @@ jobs:
     - name: install dependencies
       run: sudo apt-get --quiet install ninja-build devscripts
     - name: build.py
-      run: ./src/build.py --sync-include=cmake,wabt --build-include=wabt,debian --no-test --no-host-clang
+      run: ./src/build.py --sync-include=cmake,wabt --build-include=wabt,debian --no-test --no-host-clang --no-sysroot
--- a/src/build.py
+++ b/src/build.py
@@ -1003,6 +1003,7 @@ def Fastcomp():
       GetSrcDir('emscripten-fastcomp'),
       '-DCMAKE_CXX_FLAGS=-Wno-nonportable-include-path',
       '-DLLVM_ENABLE_LIBXML2=OFF',
+      '-DLLVM_INCLUDE_TESTS=OFF',
       '-DLLVM_INCLUDE_EXAMPLES=OFF',
       '-DLLVM_BUILD_LLVM_DYLIB=%s' % build_dylib,
       '-DLLVM_LINK_LLVM_DYLIB=%s' % build_dylib,
--- a/src/build.py
+++ b/src/build.py
@@ -85,6 +85,7 @@ LLVM_VERSION = '11.0.0'
 CLOBBER_BUILD_TAG = 15
 
 V8_BUILD_SUBDIR = os.path.join('out.gn', 'x64.release')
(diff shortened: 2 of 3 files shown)


src/build.py
```diff
@@ -818,6 +821,7 @@ def LLVM():
   command = CMakeCommandNative([
       GetLLVMSrcDir('llvm'),
       '-DCMAKE_CXX_FLAGS=-Wno-nonportable-include-path',
+      '-DLLVM_ENABLE_LIBXML2=OFF',
```
sbc100, line 824: Does features of llvm does this disable?

src/build.py
```diff
@@ -750,7 +751,9 @@ def CMakeCommandBase():
 
 def CMakeCommandNative(args, force_host_clang=True):
   command = CMakeCommandBase()
-  command.append('-DCMAKE_INSTALL_PREFIX=%s' % GetInstallDir())
+  command.extend(['-DCMAKE_INSTALL_PREFIX=%s' % GetInstallDir(),
+                  '-DCMAKE_SYSROOT=%s' %
+                  os.path.join(work_dirs.GetV8(), V8_LINUX_SYSROOT)])
```
sbc100, line 756: Is this function linux-only?

dschuff: Oops, it should only be used on Linux. fixed.

src/build.py
dschuff: As mentioned in the commit message it appears to only be used for creating Windows manifests.

.github/workflows/build.yml
```diff
@@ -33,4 +33,4 @@ jobs:
     - name: install dependencies
       run: sudo apt-get --quiet install ninja-build devscripts
     - name: build.py
-      run: ./src/build.py --sync-include=cmake,wabt --build-include=wabt,debian --no-test --no-host-clang
+      run: ./src/build.py --sync-include=cmake,wabt --build-include=wabt,debian --no-test --no-host-clang --no-sysroot
```
sbc100, line 36: Maybe add `v8` to sync-include instead?

dschuff:
This will require https://github.com/emscripten-core/emscripten-fastcomp-clang/pull/31 in order to roll into emscripten-releases.

.github/workflows/build.yml
dschuff: That will probably slow down this bot substantially. I guess the real question is, do we want to support truly-standalone use of this waterfall script or not?

dschuff: If we force use of the V8 repo with this script we can probably also remove the option to build without V8's clang.

sbc100: Would adding it to `sync` but not to `build` really slow it down that much?   I don't feel strongly about this.   Whichever you prefer.

dschuff: Yeah I don't know. probably not much, V8's DEPS aren't as bad as Chrome's. I don't really feel strongly either.
Since this is ready, I'll land it. If we want we can remove all of force_host_clang and sysroot-enabling changes together (which wouldn't really be any more work then removing force_host_clang by itself).

dschuff:
I think this is going to have to be a double-roll.

PR #637, opened 2020-04-04
Second attempt to convert python to 4-wide indentation

sbc100:
First attempt was in #617.

Again I used yapf as a base for this change, with the following manual
tweaks:

1. Manually reflow long comment lines
2. Manually indent multi-line doc strings
3. Revert some of the argparse.add_argument changes that split args onto one per line
4. Overwride yapf for certain cases that involve nestsed function
   calls.  e.g.

For example yaml seem to want to turn this:

```
    proc.check_call(SetupToolchain() +
                    ['foo', paths['win_sdk'], runtime_dirs, 'win', 'x64',
                     'environment.x64'],
                    cwd=outdir)
```

This:

```
    proc.check_call(SetupToolchain() + [
        'foo', paths['win_sdk'], runtime_dirs, 'win', 'x64', 'environment.x64'
    ],
                    cwd=outdir)
```

Choosing to go back out to column zero here in the middle of the
indented arguement list seems objectively wrong.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -801,8 +801,8 @@ def CMakeCommandNative(args):
     command.append('-DCMAKE_INSTALL_PREFIX=%s' % GetInstallDir())
     if IsLinux() and host_toolchains.ShouldUseSysroot():
         command.append('-DCMAKE_SYSROOT=%s' % GetPrebuilt(LINUX_SYSROOT))
-        command.extend('-DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++')
-        command.extend('-DCMAKE_SHARED_LINKER_FLAGS=-static-libstdc++')
+        command.append('-DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++')
+        command.append('-DCMAKE_SHARED_LINKER_FLAGS=-static-libstdc++')
 
     if host_toolchains.ShouldForceHostClang():
         command.extend(OverrideCMakeCompiler())
--- a/src/build.py
+++ b/src/build.py
@@ -41,7 +41,6 @@ import testing
 import work_dirs
 from urllib.request import urlopen, URLError
 
-
 SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
 JSVU_OUT_DIR = os.path.expanduser(os.path.join('~', '.jsvu'))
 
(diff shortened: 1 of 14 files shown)


src/build.py
```diff
@@ -276,1424 +284,1499 @@ def JavaBin():
...
+        Source('nodejs',
+               '',
+               '',  # The source and git args are ignored.
+               custom_sync=SyncPrebuiltNodeJS),
+        Source('gnuwin32',
+               '',
+               '',  # The source and git args are ignored.
+               custom_sync=SyncGNUWin32),
+        Source('wabt', GetSrcDir('wabt'), WASM_GIT_BASE + 'wabt.git'),
+        Source('binaryen', GetSrcDir('binaryen'),
```
dschuff, line 713: Why did it expand some of these and not all of them?

sbc100:
I'm not trying to go that far with this change.  I just wanted to change the intentation and this was the tool I found to do it.

Applying yapf on an ongoing basis or applying 100% are both outside the scope of what I'm trying to do here.   I don't think we need think we have to be wedded to yapf output that.   With that in mind of you have any change you'd like to make to this PR just say and I'll make them.  I don't think we have to be wedded to yapf output.

This doesn't replace flake8 and indeed that are some change that yapf tried to make that I have to revert because they cause flake8 errors :)

src/build.py
sbc100: I'dd take a look and see if I can come up with something better.  I don't know all of the yapf rules that cause this kind of thing.

dschuff:
I see, I guess I misunderstood. I think my personal preference then would be to revert the parts where it puts one argument per line everywhere.

Having said all that, I'd honestly be happy enough if we just did pick a style of "whatever this particular tool says" if we like it, because that's pretty convenient.

sbc100:
yapf doesn't seem quite ready for ths.. since it produces some code that fails flake8.

src/build.py
```diff
@@ -276,1424 +284,1499 @@ def JavaBin():
...
+        Source('fastcomp', GetSrcDir('emscripten-fastcomp'),
+               EMSCRIPTEN_GIT_BASE + 'emscripten-fastcomp.git'),
+        Source('fastcomp-clang', GetSrcDir('emscripten-fastcomp-clang'),
+               EMSCRIPTEN_GIT_BASE + 'emscripten-fastcomp-clang.git'),
+        Source('gcc',
+               GetSrcDir('gcc'),
+               GIT_MIRROR_BASE + 'chromiumos/third_party/gcc.git',
+               checkout=GCC_REVISION,
+               depth=GCC_CLONE_DEPTH),
+        Source('v8',
```
dschuff, line 692: Is it possible to put these arguments back the way they were without going over 80 cols?

src/build.py
```diff
@@ -276,1424 +284,1499 @@ def JavaBin():
...
-      self.custom_sync(self.name, self.src_dir, self.git_repo)
-    else:
-      self.GitCloneFetchCheckout()
-
-  def GitCloneFetchCheckout(self):
-    """Clone a git repo if not already cloned, then fetch and checkout."""
-    if os.path.isdir(self.src_dir):
-      print('%s directory already exists' % self.name)
+    """Metadata about a sync-able source repo on the waterfall"""
+    def __init__(self,
```
dschuff, line 448: Here

src/build.py
```diff
@@ -276,1424 +284,1499 @@ def JavaBin():
...
+
+    def PrintGitStatus(self):
+        """"Print the current git status for the sync target."""
+        print('<<<<<<<<<< STATUS FOR', self.name, '>>>>>>>>>>')
+        if os.path.exists(self.src_dir):
+            proc.check_call(['git', 'status'], cwd=self.src_dir)
+        print()
+
+
+def ChromiumFetchSync(name,
```
dschuff, line 539: Here

src/build.py
```diff
@@ -276,1424 +284,1499 @@ def JavaBin():
...
+        'llvm-dwarfdump'
+    ])
+    for p in [
+            glob.glob(os.path.join(build_dir, 'bin', b)) for b in extra_bins
+    ]:
+        for e in p:
+            CopyBinaryToArchive(os.path.join(build_dir, 'bin', e), prefix)
+
+
+def BuildEnv(build_dir,
```
dschuff, line 876: here

src/build.py
```diff
@@ -276,1424 +284,1499 @@ def JavaBin():
...
+        exclusions=LLVM_TORTURE_EXCLUSIONS,
+        out=outdir,
+        config='emscripten',
+        opt=opt)
+
+    if 0 != unexpected_result_count:
+        buildbot.Fail()
+
+
+def LinkLLVMTorture(name,
```
dschuff, line 1402: here

src/build.py
```diff
@@ -276,1424 +284,1499 @@ def JavaBin():
...
+                                                      files=input_pattern,
+                                                      fails=fails,
+                                                      attributes=[opt],
+                                                      out=outdir,
+                                                      args=args)
+    if 0 != unexpected_result_count:
+        buildbot.Fail()
+
+
+def ExecuteLLVMTorture(name,
```
dschuff, line 1425: here

src/build.py
```diff
@@ -276,1424 +284,1499 @@ def JavaBin():
...
-    self.args = args
-    self.kwargs = kwargs
-
-  def Run(self):
-    if self.os_filter and not self.os_filter.Check(BuilderPlatformName()):
-      print("Skipping %s: Doesn't work on %s" % (self.runnable.__name__,
-                                                 BuilderPlatformName()))
-      return
-    self.runnable(*self.args, **self.kwargs)
+    def __init__(self,
```
dschuff, line 1473: here

src/build.py
sbc100: Reverted

PR #638, opened 2020-04-07
Link libstdc++ statically

dschuff:
This prevents any dependence on new symbol versions in the dylib.


dschuff:
I discovered that Chrome's clang actually comes with a copy of the old libstdc++ that the bots use. In principle we should also be able to do `-Wl,-rpath-link,/path/to/old/lib` which should make the linker use the old libstdc++ instead of the system or sysroot one but for some reason it doesn't seem to work.

PR #640, opened 2020-05-07
Reland limited install of emscripten

sbc100:


src/build.py
```diff
@@ -1646,12 +1648,24 @@ def TestEmwasm():
 def ExecuteEmscriptenTestSuite(name, tests, config, outdir, warn_only=False):
     buildbot.Step('Execute emscripten testsuite (%s)' % name)
     Mkdir(outdir)
+
+    # Before we can run the tests we prepare the installed emscripten
```
dschuff, line 1652: Is `npm ci` needed before we run *any* tests (i.e. before we use emcc at all), e.g. the torture tests? Or just for the emscripten test suite?

sbc100: Yes, you are right.  In theory it is.   In practice I think we have been getting away with due to the fact that the torture tests are not using the part of emscripten that end up depending on the npm modules (i.e. closure compiler).  At least that is my guess. 

I guess we should really move this to some kind of `PrepareEmscripten` phase?  That happens after archive creation but before any tests run.   I'll leave that for a separate PR though I think

dschuff:
Yeah this seems fine for now. I don't think we have any current plans to ramp up the torture testing for now.

sbc100:
This broke the builder due to missing:  site/build/text/docs/tools_reference/emcc.txt.   Working on a fix

PR #642, opened 2020-05-09
Include all libraries in pre-built form

sbc100:
The total size of the cached libraries is 33Mb and 21Mb (uncompressed)
for wasm and fastcomp respectively so I don't think its worth being
selective here.  We should probably look at shipping LTO and PIC libs
too, although maybe we should wait until we ditch fastcomp.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1136,21 +1136,12 @@ def Emscripten(variant):
         with open(outfile, 'w') as config:
             config.write(text)
 
-    # We don't bundle every single file in the cache, since there are a lot and
-    # it would bloat our downloads. But some are useful to bundle:
-    #   * libc is large and slow to build (many tiny files).
-    #   * generated_struct_info.json is generated in a boostrapping operation
-    #     that is confusing to debug if the user has a broken setup.
-    #   * optimizer.2.exe takes a while to build and also requires a local
-    #     compiler environment for native executables.
     configs = {
-        'fastcomp': (GetInstallDir(EMSCRIPTEN_CONFIG_FASTCOMP), 'asmjs',
-                     ('libc.bc', 'generated_struct_info.json')),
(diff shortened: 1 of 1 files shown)


kripken:
20-30MB isn't a lot when shipping LLVM and such, but most of these libs would never be used by most users, so it's wasteful, especially variations like `libemmalloc-mt-debug-tracing, libgl-mt-emu-webgl2-ofb-full_es3`. Personally I'd prefer a list of the common ones. But I don't feel strongly, and I agree it's simpler this way.

sbc100:
I really like the idea of not trying to maintain some kind of whitelist and just shipping everything here since its not an unreasonable size.

It opens up some interesting possibilities.    I'm not saying we should do these right now or ever but we could consider them:

1. Remove the "build-on-demand" system completely.   Libraries are built *only* with embuilders.
2. Stop shipping the sources for system libraries to emsdk users.   Its pretty strange for an SDK to ship libc source code.  Shipping pre-built binaries is the norm.  (If you need to hack on libc itself its easy enough to grab emscripten from git).

sbc100:
I think the simplicity of not maintains a whitelist outweights that size wins.

If we want to reduce the size of the emsdk download there is plenty of other low hanging fruit to go after.   e.g. Do we ship binaryen binaries that emscripten doesn't need?    I just saved a lot of space by removing tests/third_party from emsdk .. and we could go further there.

kripken:
Sounds reasonable, I don't object. And those are interesting long-term plans I agree.

dschuff:
LGTM too.
Honestly I would also like to move to something that looks more like a traditional sysroot, (which would mean prebuilt libraries and lib/include paths that are more like the Linux FHS). 
I don't see any clear benefit for our users from the auto-build system.
I do see clear benefit for us, when we test out new compiler or library builds (it's nicer to just build what we need to run the test we want to run, instead of everything). But we'd want to be sure it's worthwhile to keep the auto-building logic just for that.

PR #643, opened 2020-05-11
Support new default emscripten cache directory

sbc100:
See https://github.com/emscripten-core/emscripten/pull/11126

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1170,7 +1170,7 @@ def Emscripten(variant):
     # Copy prebuilt/cache libraries so users don't need build them.
     Remove(GetInstallDir('lib', cache_subdir))
 
-    # Newer version of emscripten use an in-tree cache by default.
+    # Newer versions of emscripten use an in-tree cache by default.
     # TODO(sbc): Remove support for old EMSCRIPTEN_CACHE_DIR in $HOME
     in_tree_cache = GetInstallDir('emscripten', 'cache')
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -1169,9 +1169,20 @@ def WriteEmscriptenConfig(infile, outfile):
 
     # Copy prebuilt/cache libraries so users don't need build them.
     Remove(GetInstallDir('lib', cache_subdir))
-    shutil.copytree(os.path.join(EMSCRIPTEN_CACHE_DIR, cache_subdir),
+
+    # Newer version of emscripten use an in-tree cache by default.
```
kripken, line 1173: `version => versions`

src/build.py
```diff
@@ -1169,9 +1169,20 @@ def WriteEmscriptenConfig(infile, outfile):
...
+    in_tree_cache = GetInstallDir('emscripten', 'cache')
+    if os.path.exists(in_tree_cache):
+        cache_root = in_tree_cache
+    else:
+        cache_root = EMSCRIPTEN_CACHE_DIR
+
+    shutil.copytree(os.path.join(cache_root, cache_subdir),
                     GetInstallDir('lib', cache_subdir))
 
+    Remove(in_tree_cache)
```
kripken, line 1184: will this not fail if the argument does not exist?

dschuff: No, I think our `Remove` accounts for that.

PR #644, opened 2020-05-13
Ship emscripten library cache in-place

sbc100:
Now that emscripten uses an in-tree cache just ship it in place.
This also means the emsdk doesn't have to copy any of this during
post-install.


kripken:
If somehow we weren't copying it in the new place, is the emsdk the only place this would get tested I guess?

sbc100:
Yes, emsdk CI is that only place we ensure this works I believe.

This change will break the emsdk CI when the next release is made unless we land the emsdk-side of this change too: https://github.com/emscripten-core/emsdk/pull/501.  This emsdk change allows the cache to be used in-place rather than copied around.

PR #645, opened 2020-05-20
Disable libwasm in wabt build

dschuff:
Goes with WebAssembly/wabt#1435

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1023,7 +1023,8 @@ def Wabt():
     Mkdir(out_dir)
     cc_env = BuildEnv(out_dir)
 
-    cmd = CMakeCommandNative([GetSrcDir('wabt'), '-DBUILD_TESTS=OFF'])
+    cmd = CMakeCommandNative([GetSrcDir('wabt'),
+                              '-DBUILD_TESTS=OFF', '-DBUILD_LIBWASM=OFF'])
     proc.check_call(cmd, cwd=out_dir, env=cc_env)
 
     proc.check_call(['ninja', '-v'], cwd=out_dir, env=cc_env)


sbc100:
Seems sad that we have to contort our projects like this because of the limitation of gclient.

dschuff:
If we really cared we could flatten wabt's submodules into emscripten-releases' DEPS, or add a DEPS file to wabt that we could recurse into from escripten-releases. Or just switch to git submodules instead of DEPS at all (and then do our own autoroller). But we don't really need the wabt interpreter in the toolchain anyway.

sbc100:
Now that you mention it why are we including wabt here at all?

binji:
@sbc100 Probably just for a few useful tools that don't have Binaryen equivalents?

sbc100:
But does it need to be part of emsdk.. I'm pretty emscripten has no dependency on wabt.  Can we remove it from https://source.chromium.org/chromium/chromium/tools/build/+/master:scripts/slave/recipes/emscripten_releases.py?q=file:recipes%2Femscripten_releases.py&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2Fsearch%2F?

@dschuff ?

kripken:
We may want wabt's wasm2c to be a dependency soon actually!

sbc100:
Maybe... regardless we should limit emsdk to include only what emscripten needs.. it seems that right now we are shipping more than needed to all our users.

dschuff:
It does not actually need to be part of emsdk. Technically speaking for the upstream compiler we also don't need to ship any of the LLVM binutils either; but we do, because they are often useful to toolchain users. This seems similar to me?

sbc100:
If we care about download side for emsdk we should not be shipping binaries that emscripten doesn't depend on, right?    We can probably save a couple of hundred Mb limiting what we include.

dschuff:
well its just a matter of trading off the utility against the cost of the package being bigger. The tests are almost certainly not useful to users, so they should go. The static libraries could potentially be useful to someone who wants to build their own toolchain but aren't worth it, so they are not included either. Binutils are definitely useful to a few users, so it's just a question of how much it's worth to have them. Because we use the libLLVM dylib, the cost of the extra LLVM tools is quite small. But I've not actually gone and looked at what the size is.

sbc100:
Several of the binutils (-ar, -nm, -ranlib) are used directly by emscripten.

What tests do we ship?  I would hope non of the llvm tests get included by `make install`.

Anyway, wabt should be low-hanging fruit since is not used wherewhere.  Same with v8.  We do use d8 as part of emscripten testing but we pull it in via `jsvu`.    Can we at least remove those two (and avoid building them at all)?

kripken:
Again, what about keeping wabt for wasm2c? (If we think it's such a niche case people can install wabt for it manually that might be ok.)

sbc100:
Sure, if wasm2c becomes a dependency of emscripten we should keep it.   For now isn't it more like wasmer or wasmtime though where its a way of running the resulting output, and we most likely just be used in the test suite?  (like wasmtime/wasmer are).   For tools like that I think separate installation is OK.

kripken:
I'm not sure, but I think wasm2c might be used in compiling. I am thinking about maybe a flag `WASM2C` which when set `emcc` will compile using wasm2c and emit a final `.c` file containing that + the emscripten main and runtime support.

sbc100:
I agree if that becomes use useful things for emsdk users to do than we would want to include it in the emsdk.

dschuff:
> Several of the binutils (-ar, -nm, -ranlib) are used directly by emscripten. 

Sure, but I was thinking more specifically of `objdump`, `cxxfilt`, `strings`, and a bunch of others that emscripten doesn't depend on directly, but I would be sad if I had to use a toolchain without them (even as a user). I would bet money that if we removed them, someone would complain. Wabt tools arguably fall into this category too.

> What tests do we ship? I would hope non of the llvm tests get included by `make install`.

None AFAIK. I was more referring to your addition of the install script that leaves out the emscripten tests (replacing our previous method of `cp -r` as installer). That's just pure uncontroversial win.

> Anyway, wabt should be low-hanging fruit since is not used wherewhere. Same with v8. We do use d8 as part of emscripten testing but we pull it in via jsvu. Can we at least remove those two (and avoid building them at all)?

I'm fine with not shipping d8 (I'm not sure we even ship it now, do we?). But we do need to keep it in DEPS because we use its DEPS recursively to pull in our host toolchains (and I don't know about you but I am not in any hurry to reimplement all of that chrome-infra-handling stuff).

> I'm not sure, but I think wasm2c might be used in compiling. I am thinking about maybe a flag WASM2C which when set emcc will compile using wasm2c and emit a final .c file containing that + the emscripten main and runtime support.

Using wasm2c for our own purposes is very different than committing to support yet another magical feature for no (currently) known use case. We should at the very least iterate on our own use of it before we consider exposing it as part of emscripten.

kripken:
> Using wasm2c for our own purposes is very different than committing to support yet another magical feature for no (currently) known use case. We should at the very least iterate on our own use of it before we consider exposing it as part of emscripten.

sgtm. I'm just trying to avoid churn of us dropping wabt now and re-adding it in the near future. But probably fine either way.

sbc100:
OK.. lets start by not building or shipping v8.   I can't actually see where emscripten-releases specified the --build targets.     I kind of expected it to be here:

https://source.chromium.org/chromium/chromium/tools/build/+/master:scripts/slave/recipes/emscripten_releases.py;l=54?q=emscripten-releases&sq=&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2F

But I don't see anything listing build targets.

kripken:
Btw, the current wabt build wouldn't work for emscripten anyhow, as we need not just wasm2c but also the wasm2c directory with helper stuff. So we'll need something more general for that integration anyhow, and don't need to keep shipping it in the emsdk.

dschuff:
The build steps are controlled not from the recipe but from a listing in bots.json in the emscripten-releases repo: https://chromium.googlesource.com/emscripten-releases/+/refs/heads/master/bots.json#13
But before we remove that step we'll need to stop copying it to the archive: https://github.com/WebAssembly/waterfall/blob/master/src/build.py#L976

sbc100:
I don't understand, surely removing `--build-include=v8` will remove the whole step so that line won't execute?

dschuff:
oh right, nevermind. yeah we can just remove the whole build.py invocation

PR #646, opened 2020-05-30
Update expectations after wasi-libc added signal emulation

sbc100:
See https://github.com/WebAssembly/wasi-libc/pull/183

### DIFF
--- a/src/compile_torture_tests.py
+++ b/src/compile_torture_tests.py
@@ -77,8 +77,10 @@ def run(cc, cxx, testsuite, sysroot_dir, fails, exclusions, out, config, opt):
     pre_js = os.path.join(script_dir, 'em_pre.js')
 
     cflags_common = [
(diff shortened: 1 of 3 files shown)


sbc100:
Landing TBR to green the waterfall.

tlively:
If a preprocessor define and a linker flag are necessary to use the signal emulation, why did the waterfall break?

sbc100:
The think that test was failing to compile because the signal macros were not previously defined.  With this change they are defined (even without the _WASI_EMULATED_SIGNAL).

I think _WASI_EMULATED_SIGNAL guards the declaration of the `signal` function, but not the individual signal macros themselves.

sunfishcode:
_WASI_EMULATED_SIGNAL does guard the signal macros too. Possibly the difference is that previously the signal macros were defined by default (though not very useful) and now they're not.

sbc100:
Yup that sounds like it.    Without adding the macro, this change removed the signal definitions.

PR #647, opened 2020-06-02
Use EM_CONFIG when runnings emscripten tests

sbc100:
Rather than relying on the fact that tools/shared.py peeks into
sys.argv.  This change is mostly for consistency with other places
in this file such as where we run `embuilder`.


dschuff:
Consistency is good but environment variables are bad, especially on CI where we have to figure out what's going on (and reproduce it) without direct access to the execution environment. Should we have something in the logs where we show the environment, or maybe print a log when we modify it?

PR #650, opened 2020-06-11
Inject EMTEST_LACKS_NATIVE_CLANG into env for emscripten tests on bots

dschuff:
The bots don't have the proper environment vars that are needed.


dschuff:
Going to land this so it can roll before https://github.com/emscripten-core/emscripten/pull/11393 lands

PR #651, opened 2020-06-16
Add --force to git fetch

sbc100:
Without this tags that get moved will cause the fetch to fail
with `(would clobber existing tag)`

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -487,7 +487,8 @@ class Source(object):
             proc.check_call(['git'] + clone)
 
         GitUpdateRemote(self.src_dir, self.git_repo, WATERFALL_REMOTE)
-        proc.check_call(['git', 'fetch', '--tags', WATERFALL_REMOTE],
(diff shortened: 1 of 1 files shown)


dschuff:
This would clobber local changes too, right? We should maybe condition the `--force` on the runner being a bot.

sbc100:
> This would clobber local changes too, right? We should maybe condition the `--force` on the runner being a bot.

What kind of changes are you talking about?   The only local changes that could be clobber would be a if the local user modified that tag that that exists on the server... in that case the tag would be updated.  Other than that it only changes the remove branches and tags, right?

dschuff:
oh right, this is fetch and not checkout. So no effect on the working tree.
what about branches where e.g. the local branch has more changes than the remote one? Would this change the branch head and leave the changes orphaned?

sbc100:
fetch only effect with `remotes/_waterfal/xxxx` branches.. and even without `--force` it will clobber those, right?

dschuff:
yeah I think you are right.

PR #652, opened 2020-06-22
Update to Node LTS 12.18.1

dschuff:
This version has V8 v7.8, and in particular includes support for the wasm
fence instruction.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -145,7 +145,7 @@ def WindowsFSEscape(path):
 
 
 # Use prebuilt Node.js because the buildbots don't have node preinstalled
-NODE_VERSION = '12.9.1'
+NODE_VERSION = '12.18.1'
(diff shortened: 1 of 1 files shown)


kripken:
This is a different node version than the emsdk installs by default, which seems like a downside?

sbc100:
I'm ok with landing this if we agree to also update the emsdk version to 12.18.1.  (There is a script in emsdk/scripts which should make this easy enough).

dschuff:
Ah right, that's my intention, to also update emsdk.

dschuff:
If I land this, it will test this version on chromium CI. will emscripten CI also pick that up, or does it use emsdk for that?

sbc100:
I think this change will only effect the waterfall itself, not emscripten, or emsdk.

dschuff:
Ok, makes sense. I'll go ahead and land it, and we can make sure at least those tests are unaffected.

PR #653, opened 2020-06-23
Set relative symlink to XCode SDK sysroot on Mac

simonque:
This creates a symlink to the XCode SDK sysroot under /Library.
The symlink is located in the build root. This change allows the
build to use a relative sysroot path.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -788,7 +788,7 @@ def CMakeCommandBase():
     command.append('-DCMAKE_EXPORT_COMPILE_COMMANDS=ON')
     command.append('-DCMAKE_BUILD_TYPE=Release')
     if IsMac():
-        # Target macOS Seirra (10.12)
+        # Target MacOS Sierra (10.12)
         command.append('-DCMAKE_OSX_DEPLOYMENT_TARGET=10.12')
     elif IsWindows():
         # CMake's usual logic fails to find LUCI's git on Windows
@@ -797,13 +797,28 @@ def CMakeCommandBase():
     return command
 
 
-def CMakeCommandNative(args):
+def CMakeCommandNative(args, build_dir):
     command = CMakeCommandBase()
     command.append('-DCMAKE_INSTALL_PREFIX=%s' % GetInstallDir())
     if IsLinux() and host_toolchains.ShouldUseSysroot():
         command.append('-DCMAKE_SYSROOT=%s' % GetPrebuilt(LINUX_SYSROOT))
         command.append('-DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++')
         command.append('-DCMAKE_SHARED_LINKER_FLAGS=-static-libstdc++')
+
+    elif IsMac() and host_toolchains.ShouldUseSysroot():
+        # Get XCode SDK path.
+        xcode_sdk_path = proc.check_output(['xcrun',
+                                            '--show-sdk-path']).strip()
+        # Create relpath symlink if it doesn't exist.
+        # If it does exist, but points to a different location, update it.
+        symlink_path = os.path.join(build_dir, 'xcode_sdk')
+        if os.path.lexists(
+                symlink_path) and os.readlink(symlink_path) != xcode_sdk_path:
+            os.remove(symlink_path)
+        if not os.path.exists(symlink_path):
+            os.symlink(xcode_sdk_path, symlink_path)
+        command.append('-DCMAKE_OSX_SYSROOT=%s' % symlink_path)
+        command.append('-DCMAKE_SYSROOT=%s' % symlink_path)
 
     if host_toolchains.ShouldForceHostClang():
         command.extend(OverrideCMakeCompiler())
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -788,23 +791,35 @@ def CMakeCommandBase():
...
         command.append('-DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++')
         command.append('-DCMAKE_SHARED_LINKER_FLAGS=-static-libstdc++')
 
+    elif IsMac() and host_toolchains.ShouldUseSysroot():
+        # Create relpath symlink.
+        print("Creating symlink to %s" % os.path.join(build_dir, 'xcode_sdk'))
+        symlink_dest = os.path.join(build_dir, 'xcode_sdk')
+        if not os.path.exists(symlink_dest):
+            print("creating symlink %s", symlink_dest)
+            os.symlink('/Library/Developer/CommandLineTools/SDKs', symlink_dest)
```
simonque, line 816: I wasn't sure if there was a programmatic way to get '/Library/Developer/CommandLineTools/SDKs' which is why it's hard-coded here.

If there's a better way to do this let me know.

simonque: BTW the emscripten-releases builder also has  the path '/Library/Developer/CommandLineTools/SDKs'.

src/build.py
```diff
@@ -788,23 +791,37 @@ def CMakeCommandBase():
...
         command.append('-DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++')
         command.append('-DCMAKE_SHARED_LINKER_FLAGS=-static-libstdc++')
 
+    elif IsMac() and host_toolchains.ShouldUseSysroot():
+        # Create relpath symlink.
+        print("Creating symlink to %s" % os.path.join(build_dir, 'xcode_sdk'))
+        symlink_dest = os.path.join(build_dir, 'xcode_sdk')
+        if not os.path.exists(symlink_dest):
+            print("creating symlink %s", symlink_dest)
+            os.symlink('/Library/Developer/CommandLineTools/SDKs',
```
dschuff, line 816: can we move this hardcoded path up next to OSX_TARGET_VERSION?

src/build.py
```diff
@@ -788,23 +791,37 @@ def CMakeCommandBase():
...
-def CMakeCommandNative(args):
+def CMakeCommandNative(args, build_dir):
     command = CMakeCommandBase()
     command.append('-DCMAKE_INSTALL_PREFIX=%s' % GetInstallDir())
     if IsLinux() and host_toolchains.ShouldUseSysroot():
         command.append('-DCMAKE_SYSROOT=%s' % GetPrebuilt(LINUX_SYSROOT))
         command.append('-DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++')
         command.append('-DCMAKE_SHARED_LINKER_FLAGS=-static-libstdc++')
 
+    elif IsMac() and host_toolchains.ShouldUseSysroot():
```
dschuff, line 810: Don't mac builds basically always use a sysroot, that comes with the SDK? What would !ShouldUseSysroot mean?

simonque: I just copied it from the IsLinux case. I can remove the ShouldUseSysroot().

src/build.py
simonque: OK

simonque:
Updated with changes.

src/build.py
dschuff: FWIW my mac laptop does not have this path. The command line tools must be located somewhere else on my machine.

src/build.py
dschuff: Actually since we're passing this explicit flag to CMake we should probably keep it behind `ShouldUseSysroot`. Mostly I'm still trying to fully  understand what this change actually does.

src/build.py
```diff
@@ -788,23 +793,32 @@ def CMakeCommandBase():
...
 
+    elif IsMac():
+        # Create relpath symlink.
+        symlink_dest = os.path.join(build_dir, 'xcode_sdk')
+        if not os.path.exists(symlink_dest):
+            os.symlink(OSX_SDKS_PATH, symlink_dest)
+        xcode_sdk_path = os.path.join(symlink_dest,
+                                      'MacOSX%s.sdk' % OSX_TARGET_VERSION)
+        command.append('-DCMAKE_OSX_SYSROOT=%s' % xcode_sdk_path)
+        command.append('-DCMAKE_SYSROOT=%s' % xcode_sdk_path)
```
dschuff, line 820: do we actually need `DCMAKE_SYSROOT` if we already also have `DCMAKE_OSX_SYSROOT`?

src/build.py
simonque: In crbug.com/1070150, the Mac build fails due to an absolute path for --sysroot, which is not supported by the Goma backend. To work around that, we create a symlink under the build root to the actual sysroot path, so that it looks like a relative path. We pass in args to cmake to set up the build to use that relative sysroot path.

src/build.py
simonque: Ideally, there should be some way to determine how Cmake determines the sysroot on Mac.

dschuff: OK, I think I have it:
`xcrun --show-sdk-path` prints the path to the current SDK. On my machine that's
'/Users/dschuff/code/osx_sdk/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk'
So we can just run that to find the SDKs path.

PR #654, opened 2020-07-05
Run `npm ci` before trying to use emscripten

sbc100:
We recently started depending more on the result on `npm ci` because
we now get acorn compiler via npm.


dschuff:
LGTM. It looks like this mostly affects the tests like `TestEmWasm` that are run on the wasm-stat.us waterfall rather than emscripten-releases, correct?

sbc100:
Yup.. the addition of acorn dependency broke the wasm-stat.us  waterfall.

PR #656, opened 2020-07-17
Remove sanity.txt after populating the emscripten cache

sbc100:


src/build.py
```diff
@@ -1181,6 +1181,12 @@ def WriteEmscriptenConfig(infile, outfile):
     finally:
         del os.environ['EM_CONFIG']
 
+    # Remove the santiy file.  This means its will get re-generated on first
```
kripken, line 1184:
```suggestion
    # Remove the sanity file.  This means it will get generated on first
```

PR #658, opened 2020-07-29
Cleanup emscripten config files

sbc100:
These files are not used by the emsdk but only internally
when testing the resulting SDK.

### DIFF
--- a/src/emscripten_config_fastcomp
+++ b/src/emscripten_config_fastcomp
@@ -1,62 +1,19 @@
-
-# Note: If you put paths relative to the home directory, do not forget os.path.expanduser
-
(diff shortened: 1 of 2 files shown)


dschuff:
So the change here is just to remove all the boilerplate and text meant to help users write their own emscripten configs?

As an aside: the current way we run the tests is to have these files sitting in the install directory, and then we pass --em-config manually on all of our emscripten-releases test runs (but that was before we had embedded config capability). Also IIRC they get packaged up into the release package unnecessarily.
Should we change it so that we create the release package first, then just copy this as an embedded config for testing?
(or perhaps do that after we stop testing fastcomp so there's only one?)

sbc100:
We could do that as followup yes.

PR #663, opened 2020-09-08
Stop building asmjs optimizer

dschuff:
It's been removed from emscripten


dschuff:
nevermind this already landed, it's just not mirrored yet.

PR #664, opened 2020-09-09
Switch builders to use builder_group property instead of mastername.

dschuff:
Bug: 1109276
Change-Id: I1d0d70d6680d0252c72700a4a03fc4c309eda63a


dschuff:
Patch mirrored from the gerrit CL by @kleerwater

PR #665, opened 2020-09-10
Remove fastcomp support

sbc100:


src/build.py
```diff
@@ -1483,9 +1430,7 @@ def AllBuilds():
         Build('jsvu', Jsvu, os_filter=Filter(exclude=['windows'])),
         Build('wabt', Wabt),
         Build('binaryen', Binaryen),
-        Build('fastcomp', Fastcomp),
-        Build('emscripten-upstream', Emscripten, None, True, 'upstream'),
-        Build('emscripten-fastcomp', Emscripten, None, True, 'fastcomp'),
+        Build('emscripten', Emscripten),
```
dschuff, line 1433: Just renaming `emscripten-upstream` to `emscripten` is going to break unless we change the emscripten-releases builder (https://chromium.googlesource.com/emscripten-releases/+/refs/heads/master/bots.json) simultaneously with the roll, or just allow both names.

dschuff (approved):
I don't particularly care which way we get the roll in; I'm sheriff this week, I can probably just do a roll + update of the bot command.

src/build.py
sbc100: OK, I'll revert that part for now

kripken (approved):

PR #670, opened 2020-10-09
Include node_modules in uploaded bundle

sbc100:
This should allow the bundle to be used directly without running
`npm install` or `npm ci` as post-install step.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1077,7 +1077,7 @@ def InstallEmscripten():
     #   2. A bug with the native compiler has bloated the packages from
     #      30MB to almost 300MB
     #      https://github.com/google/closure-compiler-npm/issues/186
-    # If either of these bugs are fixed we could consdider removing this
+    # If either of these bugs are fixed we could consider removing this
     # hack.
     native = None
     if IsMac():
--- a/src/build.py
+++ b/src/build.py
@@ -1065,6 +1065,29 @@ def InstallEmscripten():
     proc.check_call([os.path.join('tools', 'install.py'), em_install_dir],
                     cwd=src_dir)
 
+    print('Running npm install ...')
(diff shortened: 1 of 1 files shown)


sbc100:
@pfaffe

src/build.py
```diff
@@ -1065,6 +1065,29 @@ def InstallEmscripten():
...
+    # Manually install the appropriate native Closure Compiler package
+    # This is currently needed because npm ci will install the packages
+    # for Closure for all platforms, adding 180MB to the download size
+    # There are two problems here:
+    #   1. npm ci does not consider the platform of optional dependencies
+    #      https://github.com/npm/cli/issues/558
+    #   2. A bug with the native compiler has bloated the packages from
+    #      30MB to almost 300MB
+    #      https://github.com/google/closure-compiler-npm/issues/186
+    # If either of these bugs are fixed we could consdider removing this
```
kripken, line 1080:
```suggestion
    # If either of these bugs are fixed we could consider removing this
```

kripken (commented):
Sounds like this increases package size by 30MB?

Do the packages still work if the user does `npm ci`?

I see the `npm ci` command here, but where are the `node_modules` files copied? Or do we copy a parent directory with all subchildren?

sbc100:
> Sounds like this increases package size by 30MB?
> 

Yes, but its saves each user from having to run `npm ci` on install .. which in turn saves at least (probably a fair bit more) in overall downloading,

> Do the packages still work if the user does `npm ci`?

Yes, I believe users can always run this, even if we pre-seed the directory modules.

> 
> I see the `npm ci` command here, but where are the `node_modules` files copied? Or do we copy a parent directory with all subchildren?

The the command is run in `em_install_dir` which is the installed directed that is ultimately tar'd up.

kripken (approved):

dschuff (approved):

PR #671, opened 2020-10-09
Remove some unused args

sbc100:
This fixes what looks like an actual bug in make_blocking


dschuff (commented):
Are these 2 sets of changes related?

Also, is the mac issue still happening? maybe we can just remove make_blocking

sbc100:
Its one set of changes that was generated by a "detect unused arguments" flake8 check.

It just happened that one of the used arguments represented a bug.  Yes we can probably remove that block .. maybe a separate change though.

dschuff (approved):

PR #673, opened 2020-10-24
Use custom sysroot creator to build our linux sysroot

sbc100:
Landing initially with debian/sarge which means nothing
should change/break.  The plan being to update to something
with a more recent libc++.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,6 @@ jobs:
     - name: install dependencies
       run: sudo apt-get --quiet install ninja-build devscripts
     - name: build.py
-      run: ./src/build.py --sync-include=sysroot,cmake,wabt --build-include=wabt,debian --no-test --no-host-clang --no-sysroot
+      run: ./src/build.py --sync-include=cmake,wabt --build-include=wabt,debian --no-test --no-host-clang --no-sysroot
     - name: build.py with sysroot
       run: ./src/build.py --sync-include=sysroot,cmake,binaryen --build-include=binaryen --no-test --no-host-clang
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,4 +33,6 @@ jobs:
     - name: install dependencies
       run: sudo apt-get --quiet install ninja-build devscripts
     - name: build.py
-      run: ./src/build.py --sync-include=cmake,wabt --build-include=wabt,debian --no-test --no-host-clang --no-sysroot
+      run: ./src/build.py --sync-include=sysroot,cmake,wabt --build-include=wabt,debian --no-test --no-host-clang --no-sysroot
+    - name: build.py with sysroot
+      run: ./src/build.py --sync-include=sysroot,cmake,binaryen --build-include=binaryen --no-test --no-host-clang
--- a/src/create_sysroot.sh
+++ b/src/create_sysroot.sh
@@ -27,6 +27,6 @@ for d in dev proc tmp home run var boot media sys srv mnt; do
   rm -rf $TARGET_DIR/$d
 done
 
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -33,4 +33,6 @@ jobs:
     - name: install dependencies
       run: sudo apt-get --quiet install ninja-build devscripts
     - name: build.py
-      run: ./src/build.py --sync-include=cmake,wabt --build-include=wabt,debian --no-test --no-host-clang --no-sysroot
+      run: ./src/build.py --sync-include=sysroot,cmake,wabt --build-include=wabt,debian --no-test --no-host-clang --no-sysroot
```
dschuff, line 36: do we want sysroot in the sync if it builds with --no-sysroot?

src/create_sysroot.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# Create a linux sysroot image (tar archive) based on debian
+# This is than used in the build scripts as the --sysroot arguments
+# when compiling linux binaries.
+#
+# See https://xw.is/wiki/Create_Debian_sysroots
+
+set -e
```
dschuff, line 9: I prefer using the `-o` flags to set (i.e. `set -o errexit` in this case) since it's much more obvious what they do.

src/create_sysroot.sh
```diff
@@ -0,0 +1,32 @@
...
+
+SUITE=stretch
+TARGET_DIR=sysroot_debian_${SUITE}_amd64
```
dschuff, line 12: if we plan to do more than one version of the sysroot, maybe it should get a number too?

src/create_sysroot.sh
```diff
@@ -0,0 +1,32 @@
...
+find $TARGET_DIR -type l -lname '/*' -exec sh -c 'file="$0"; dir=$(dirname "$file"); target=$(readlink "$0"); prefix=$(dirname "$dir" | sed 's@[^/]*@\.\.@g'); newtarget="$prefix$target"; echo ln -snf $newtarget $file' {} \;
+
+# Remove parts that are not relevent to --sysroot
+for d in dev proc tmp home run var boot media sys srv mnt; do
+  rm -rf $TARGET_DIR/$d
+done
+
+tar cJf sysroot_debian_${SUITE}_amd64.tar.xz -C $TARGET_DIR .
+
+gsutil cp sysroot_debian_${SUITE}_amd64.tar.xz gs://wasm/
```
dschuff, line 32: Is the plan to run this on a bot? If not, do we need it to automatically upload too?

sbc100: This is more to document how to upload it.. since uploading is only logical thing to do after running this script... but yeah.. maybe i'll move it to the comment at the top.. especially since  this scripte needs to run as root and the upload will fail in this case (at least for me).

src/create_sysroot.sh
sbc100: We an add that later if needed I think.

src/create_sysroot.sh
sbc100: Done

.github/workflows/build.yml
sbc100: Ooops.

src/create_sysroot.sh
sbc100: Done

dschuff (approved):
LGTM; we can't land it yet though because AFAICS the sysroot hasn't been uploaded yet, right?

sbc100:
No I did upload it.. hence the CI tests pass right?  They would fail if I hadn't.

dschuff:
oh, you're right; I checked but failed to scroll to the 2nd page in the list of files 🤦‍♀️

PR #675, opened 2020-10-27
Switch sysroot to ubuntu bionic

sbc100:
This is enough to allow C++-17 to be used in he binaryen build.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -77,8 +77,8 @@ CLOBBER_BUILD_TAG = 19
 
 V8_BUILD_SUBDIR = os.path.join('out.gn', 'x64.release')
 
-LINUX_SYSROOT = 'sysroot_debian_stretch_amd64'
-LINUX_SYSROOT_URL = WASM_STORAGE_BASE + LINUX_SYSROOT + '_v2.tar.xz'
+LINUX_SYSROOT = 'sysroot_ubuntu_bionic_amd64'
+LINUX_SYSROOT_URL = WASM_STORAGE_BASE + LINUX_SYSROOT + '_v1.tar.xz'
 
 options = None
 
--- a/src/create_sysroot.sh
+++ b/src/create_sysroot.sh
@@ -12,17 +12,18 @@
 
 set -o errexit
 
-SUITE=stretch
-TARGET_DIR=sysroot_debian_${SUITE}_amd64
-VERSION=2
+SUITE=bionic
+DISTRO=ubuntu
+TARGET_DIR=sysroot_${DISTRO}_${SUITE}_amd64
+VERSION=1
 
 mkdir $TARGET_DIR
 
 # Perform minimal installation
-debootstrap $SUITE $TARGET_DIR http://deb.debian.org/debian
+debootstrap $SUITE $TARGET_DIR http://archive.ubuntu.com/ubuntu/
 
 # Install additional packages
-chroot $TARGET_DIR apt-get install -y -q libstdc++-6-dev zlib1g-dev
+chroot $TARGET_DIR apt-get install -y -q libstdc++-dev zlib1g-dev
 
 # Convert absolute symlinks to relative
 find $TARGET_DIR -type l -lname '/*' -exec sh -c 'file="$0"; dir=$(dirname "$file"); target=$(readlink "$0"); prefix=$(dirname "$dir" | sed 's@[^/]*@\.\.@g'); newtarget="$prefix$target"; ln -snf $newtarget $file' {} \;
@@ -32,4 +33,4 @@ for d in dev proc tmp home run var boot media sys srv mnt; do
   rm -rf $TARGET_DIR/$d
 done
 
-tar cJf sysroot_debian_${SUITE}_amd64_v${VERSION}.tar.xz -C $TARGET_DIR .
+tar cJf sysroot_${DISTRO}_${SUITE}_amd64_v${VERSION}.tar.xz -C $TARGET_DIR .


kripken (commented):
Forgive my ignorance, but does using a fairly new distro mean that the builds will only be usable on that version and newer, and nothing older?

sbc100:
Yes, switch from stretch to bionic will slightly decrease compatibility with older machines.

However, we are looking to switch to C++17.. and bionic was released in 2018 so it seems like a pretty good match/balance:  https://wiki.ubuntu.com/Releases

One way around that regression would to statically link libc++ rather than relying on the distro, but I'm not sure its worth the effort.  Remember, this only effects users who want to use our pre-built emsdk binaries.

Actually there is that distro we use in the binaryen CI which produces completely static binaries (alpine linux) so if we find that users are effected by this we can always go that route.

sbc100:
I propose that we looking to providing a completely static alpine/linux build, but only if we have users who are effected by this transition, since it will increase overall code size a little.

(if fact I will investigate now to find out how much size is impacted).

sbc100:
I confirmed that version dependencies for libLLVM are the same both before and after this change:

```
Version References:
  required from librt.so.1:
    0x09691a75 0x00 21 GLIBC_2.2.5
  required from ld-linux-x86-64.so.2:
    0x0d696913 0x00 18 GLIBC_2.3
  required from libm.so.6:
    0x06969187 0x00 19 GLIBC_2.27
    0x09691a75 0x00 09 GLIBC_2.2.5
  required from libz.so.1:
    0x0827e5c0 0x00 08 ZLIB_1.2.0
  required from libc.so.6:
    0x06969195 0x00 20 GLIBC_2.15
    0x09691974 0x00 17 GLIBC_2.3.4
    0x0d696914 0x00 15 GLIBC_2.4
    0x0d696916 0x00 14 GLIBC_2.6
    0x0d696913 0x00 10 GLIBC_2.3
    0x06969194 0x00 07 GLIBC_2.14
    0x09691a75 0x00 06 GLIBC_2.2.5
  required from libpthread.so.0:
    0x09691972 0x00 16 GLIBC_2.3.2
    0x06969192 0x00 11 GLIBC_2.12
    0x09691a75 0x00 05 GLIBC_2.2.5
  required from libgcc_s.so.1:
    0x0b792653 0x00 13 GCC_3.3
    0x09276060 0x00 12 GCC_4.2.0
    0x0b792650 0x00 04 GCC_3.0
  required from libdl.so.2:
    0x09691a75 0x00 03 GLIBC_2.2.5
```

This means I don't think this change will have any effect on compatibility.

kripken:
> Remember, this only effects users who want to use our pre-built emsdk binaries.

Isn't that most of our users, though? But since it sounds like there is no change in practice, sgtm.

Side issue, how does that work on linux, to depends on multiple glibc versions from a single library? Or am I reading that wrong?

kripken (approved):

sbc100:
> > Remember, this only effects users who want to use our pre-built emsdk binaries.
> 
> Isn't that most of our users, though? But since it sounds like there is no change in practice, sgtm.

Yes, I guess my point is that we already have users of ancient linux distros and they have the fallback option of building from source, or using the docker image.

When we do make a change, such as the C++ update in binaryen I think the number of user will increase.  Trying to figure out in advance exactly how many users will be effected is pretty hard, since we don't really have any metrics. 

I suggest we just try and see how much feedback we get :-/

> 
> Side issue, how does that work on linux, to depends on multiple glibc versions from a single library? Or am I reading that wrong?

I'm not sure.. ELF symbol versioning is not something I'm very familiar with.

kripken:
Sounds like a reasonable plan.

May be worth adding something to the changelog to alert people.

sbc100:
That changelog entry would be required if/when we land the binaryen C++17 update I believe.

Its also not clear where such a changelog lives.. since its a binaryn change that efffects the emsdk binaries.. its really not an emscripten change at all.

dschuff:
> Side issue, how does that work on linux, to depends on multiple glibc versions from a single library? Or am I reading that wrong?

Each symbol can have its own version. So if you use a symbol that hasn't changed since glibc 2.2.5 you get a symbol with that version, but if you use one that has changed more recently, then you have a newer dependence.

kripken:
@sbc100 

> That changelog entry would be required if/when we land the binaryen C++17 update I believe.

Doesn't this immediately change how we build things, even if we don't take advantage of the benefits like c++17? (Why would actually using c++17 be noticeable, and not the dependency on the system libraries, which change now I think?)

@dschuff 

> Each symbol can have its own version. So if you use a symbol that hasn't changed since glibc 2.2.5 you get a symbol with that version, but if you use one that has changed more recently, then you have a newer dependence.

I see, thanks!

sbc100:
> @sbc100
> 
> > That changelog entry would be required if/when we land the binaryen C++17 update I believe.
> 
> Doesn't this immediately change how we build things, even if we don't take advantage of the benefits like c++17? (Why would actually using c++17 be noticeable, and not the dependency on the system libraries, which change now I think?)
> 

I'm just going by my empirical measurement of the symbol versions not changing with this change.

Maybe they won't change when we switch to C++17 but its possible that using new features will introduce new versions of the dependencies.   We will have to see I guess.   But at least it should build and run!

> @dschuff
> 
> > Each symbol can have its own version. So if you use a symbol that hasn't changed since glibc 2.2.5 you get a symbol with that version, but if you use one that has changed more recently, then you have a newer dependence.
> 
> I see, thanks!

sbc100:
And... I was wrong.

The ` 0x06969187 0x00 19 GLIBC_2.27` dep is new and causing failures on bots:

```
subprocess.check_call(`/b/s/w/ir/cache/vpython/75e09f/bin/python /b/s/w/ir/k/install/emscripten/embuilder.py build SYSTEM`, cwd=`/b/s/w/ir/cache/builder/emscripten-releases/waterfall/src`)
/b/s/w/ir/k/install/bin/clang: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27' not found (required by /b/s/w/ir/k/install/bin/../lib/libLLVM-12git.so)
embuilder: error: '/b/s/w/ir/k/install/bin/clang --version' failed (1)
```

Reverting this change

PR #679, opened 2021-01-12
Add a build.py flag to build with extra optimizations

dschuff:
This enables ThinLTO for LLVM and Binaryen and disables assertions on LLVM. It's off by default.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1062,7 +1062,6 @@ def Binaryen():
     cmake_command = CMakeCommandNative([GetSrcDir('binaryen')], out_dir)
     if options.extraopt:
         cmake_command.append('-DBYN_ENABLE_LTO=ON')
-        cmake_command.append('-DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=lld')
     proc.check_call(cmake_command,
                     cwd=out_dir,
                     env=cc_env)
--- a/src/build.py
+++ b/src/build.py
@@ -764,11 +764,14 @@ def OverrideCMakeCompiler():
         return []
     cc = 'clang-cl' if IsWindows() else 'clang'
     cxx = 'clang-cl' if IsWindows() else 'clang++'
-    return [
+    tools = [
         '-DCMAKE_C_COMPILER=' + Executable(GetPrebuiltClang(cc)),
-        '-DCMAKE_CXX_COMPILER=' + Executable(GetPrebuiltClang(cxx))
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -1785,6 +1800,9 @@ def SplitComma(arg):
     parser.add_argument(
         '--clobber', dest='clobber', default=False, action='store_true',
         help="Delete working directories, forcing a clean build")
+    parser.add_argument(
+        '--extraopt', dest='extraopt', default=False, action='store_true',
```
sbc100, line 1804: This reads to be as "extra options" so I expected to be able to pass more options this way.   I don't have any great alternatives though.   We could just called it `--enable-lto` I guess?

sbc100 (approved):

src/build.py
kripken: I also found `extraopt` slightly confusing. Maybe `enable-lto` is less confusing even though it also disables assertions. I don't have an idea for a better name.

dschuff: `enable-lto` seems fine to me. If we ever change it so that assertions are selected differently from lto we can update to whatever makes sense.

dschuff: I went with `use-lto` to match the use_sysroot option (and to provide just a little extra disambiguation between using LTO in the build and enabling LLVM's LTO target which builds libLTO)

src/build.py
```diff
@@ -764,10 +764,15 @@ def OverrideCMakeCompiler():
...
     cxx = 'clang-cl' if IsWindows() else 'clang++'
-    return [
+    tools = [
         '-DCMAKE_C_COMPILER=' + Executable(GetPrebuiltClang(cc)),
-        '-DCMAKE_CXX_COMPILER=' + Executable(GetPrebuiltClang(cxx))
+        '-DCMAKE_CXX_COMPILER=' + Executable(GetPrebuiltClang(cxx)),
     ]
+    if IsWindows():
+        tools.append('-DCMAKE_LINKER=' +
+                     Executable(GetPrebuiltClang('lld-link')))
```
kripken, line 773: why do we need this on windows specifically?

kripken (approved):
lgtm % question

src/build.py
dschuff: On Windows, the linker is apparently invoked directly by CMake rather than indirectly by the compiler driver as on other platforms. So CMake itself needs to have the path to the linker.

kripken: I see, thanks.

kripken (approved):

PR #680, opened 2021-01-15
Use LTO when building tagged revisions of emscripten

dschuff:
Adds a third mode for building with LTO: aside from enabled or disabled,
check whether the current revision of emscripten is a tagged revision,
and use LTO if so.

Also clobber working dir before and after LTO builds


kripken:
I believe this would only work if we tag first and build later, which is the opposite of what we currently do. Right now we build, see that tests pass, and then "bless" an already-built revision as a release.

PR #682, opened 2021-01-23
Disable prebuilt clang and LLD when building with LTO on mac

dschuff:
LLD isn't fully baked on mac yet, and the prebuilt clang
doesn't include libLTO. But the tools included with the
default SDK seem to work.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -816,7 +816,8 @@ def CMakeCommandNative(args, build_dir):
 
     if host_toolchains.ShouldForceHostClang():
         command.extend(OverrideCMakeCompiler())
-        # Goma doesn't have MSVC in its cache, so don't use it in this case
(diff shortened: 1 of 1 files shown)


tlively (approved):

sbc100:
So apple's linker somehow supports LTO?  Or does avoiding lld also effectively turn off LTO?

dschuff:
Apple's linker does support LTO (it's had a library called libLTO for quite
some time, which is built as part of LLVM, sort of like the gold plugin but
with a different API).
I tried this out this locally and it seems to work fine, but it's suuuuper
slow (maybe it's ignoring the "thin" part and doing full LTO?) so it won't
work on our bots even after I doubled the time limit to 4 hours. I'll keep
hammering on it to see what I can do.

PR #683, opened 2021-01-27
Limit which LLVM tools are built in release/LTO mode

dschuff:
Reducing the number of link steps significantly reduces build time, and many of the built targets are not installed or used.
This uses the `LLVM_DISTRIBUTION_COMPONENTS` mechanism as documented in the [distribution guide](https://llvm.org/docs/BuildingADistribution.html).

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -73,7 +73,7 @@ LLVM_VERSION = '12.0.0'
 # Update this number each time you want to create a clobber build.  If the
 # clobber_version.txt file in the build dir doesn't match we remove ALL work
 # dirs.  This works like a simpler version of chromium's landmine feature.
-CLOBBER_BUILD_TAG = 21
+CLOBBER_BUILD_TAG = 22
 
 V8_BUILD_SUBDIR = os.path.join('out.gn', 'x64.release')
 
--- a/src/build.py
+++ b/src/build.py
@@ -904,7 +904,6 @@ def LLVM():
         '-DCLANG_ENABLE_STATIC_ANALYZER=OFF',
     ], build_dir)
 
-
     if not IsMac():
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -902,24 +900,38 @@ def LLVM():
...
+    if not IsMac():
+        # LLD isn't fully baked on mac yet.
+        command.append('-DLLVM_ENABLE_LLD=ON')
+
+    ninja_targets = ('all', 'install')
     if options.use_lto:
+        targets = ['clang', 'lld', 'llvm-ar', 'llvm-addr2line', 'llvm-cxxfilt',
+                   'llvm-dwarfdump', 'llvm-dwp', 'llvm-nm', 'llvm-objcopy',
+                   'llvm-objdump', 'llvm-ranlib', 'llvm-readobj', 'llvm-size',
+                   'llvm-strings', 'llvm-symbolizer']
```
kripken, line 916: What do we use addr2line, cxxfilt, readobj, size, strings, and symbolizer for?

And until we remove llc from emscripten, I think we still need it?

dschuff: these are all just the standard binutils utilities, which sometimes have general usefulness for toolchain users and IMO just generally belong in a toolchain. We should add llvm-strip too, once I get around to fixing it. (addr2line is just a symlink to symbolizer).
I added 'opt' and 'llc' on a separate line to make them easier to remove later whenever we're ready.

kripken: I see, thanks. And yes, I missed the other line with opt and llc...

kripken (approved):
I'm not very familiar with LLVM's build flags, but these look right assuming it builds ok...

PR #684, opened 2021-01-28
Clobber the work dir of incremental builds on failure

dschuff:
Also clobber before and after LTO builds.

This is accomplished by hooking into the `Run` method that runs all build rules. 
Because of that, I also moved the specification of which directory is the build directory out of the body of each build function (so the build runner could access it)

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -882,9 +882,8 @@ def BuildEnv(build_dir, use_gnuwin32=False, bin_subdir=False,
     return cc_env
 
 
-def LLVM():
+def LLVM(build_dir):
     buildbot.Step('LLVM')
-    build_dir = os.path.join(work_dirs.GetBuild(), 'llvm-out')
     Mkdir(build_dir)
     cc_env = BuildEnv(build_dir, bin_subdir=True)
     build_dylib = 'ON' if not IsWindows() and not options.use_lto else 'OFF'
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -693,6 +693,12 @@ def AllSources():
     ]
 
 
+def ClobberOneDir(work_dir):
+    if buildbot.IsBot():
+        Remove(work_dir)
+        Mkdir(work_dir)
```
sbc100, line 699: No need to mkdir here I think.

src/build.py
```diff
@@ -693,6 +693,12 @@ def AllSources():
     ]
 
 
+def ClobberOneDir(work_dir):
```
sbc100, line 696: I'm not crazy about the name of this function.

How about `RemoveBuildDir` .. or `RemoveIfIsBot`?   Not great alternatives.

sbc100 (approved):

src/build.py
dschuff: Yeah the "clobber" name goes way back to the original buildbot-based system, so "clobber" kind of implies that it has meaning only on bots. `RemoveIfBot` seems fine given how simple it is.

src/build.py
```diff
@@ -1416,20 +1416,40 @@ def ValidateLLVMTorture(indir, ext, opt):
 
 class Build(object):
     def __init__(self, name_, runnable_, os_filter=None, is_default=True,
-                 *args, **kwargs):
+                 incremental_build_dir=None, *args, **kwargs):
         self.name = name_
         self.runnable = runnable_
         self.os_filter = os_filter
-        self.is_deafult = is_default
```
kripken, line 1423: This removes the only use of the `is_default` parameter, so perhaps it is not needed?

src/build.py
```diff
@@ -1416,20 +1416,40 @@ def ValidateLLVMTorture(indir, ext, opt):
...
             print("Skipping %s: Doesn't work on %s" %
                   (self.runnable.__name__, BuilderPlatformName()))
             return
-        self.runnable(*self.args, **self.kwargs)
+
+        try:
+            # When using LTO we always want a clean build (the previous
+            # build was non-LTO)
+            if self.incremental_build_dir and options.use_lto:
+                RemoveIfBot(self.incremental_build_dir)
```
kripken, line 1440: can the if be outside the try?

src/build.py
```diff
@@ -1416,20 +1416,40 @@ def ValidateLLVMTorture(indir, ext, opt):
...
+            self.runnable(*self.args, **self.kwargs)
+        except Exception:
+            # If the build fails (even non-LTO), a possible cause is a build
+            # config change, so clobber the work dir for next time.
+            if self.incremental_build_dir:
+                RemoveIfBot(self.incremental_build_dir)
+            raise
+        finally:
+            # When using LTO we want to always clean up afterward,
+            # (the next build will be non-LTO.
```
kripken, line 1450:
```suggestion
            # (the next build will be non-LTO).
```

src/build.py
dschuff: Yeah, that would make sense given that the except repeats it :)

kripken (approved):

PR #688, opened 2021-03-04
Build Binaryen with static libs when using LTO

dschuff:
This bloats the tool install size by about 30MB but also improves
speed by about 17% (using the totally-scientific metric of timing
check.py on my desktop)

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1081,6 +1081,7 @@ def Binaryen(build_dir):
 
     cmake_command = CMakeCommandNative([GetSrcDir('binaryen')], build_dir)
     if options.use_lto:
+        cmake_command.append('-DBUILD_STATIC_LIB=ON')
         cmake_command.append('-DBYN_ENABLE_LTO=ON')
 
     proc.check_call(cmake_command, cwd=build_dir, env=cc_env)


dschuff:
The other option would be to force BUILD_STATIC_LIB=ON in CMakeLists.txt whenever BYN_ENABLE_LTO is on. The advantage of that would be making it easier to build the likely-best configuration, but disallowing a perfectly valid configuration (LTO+dynamic libs, which is also the smallest overall install size).

kripken:
How much is the 30MB on top of?

Looking locally, libbinaryen is 15MB and everything else is pretty small, so I am guessing this is a big increase percentage-wise?

kripken:
(but, I assume, a small increase percentage wise for the entire waterfall download)

kripken (approved):
lgtm assuming that's right.

Might be good measuring on something more realistic. 17% of check.py is probably an understatement since there is a lot of fixed overhead there. Doing `wasm-opt -O3` on a big wasm file (like poppler from the test suite) could be useful.

dschuff:
Yes, this is compared to 6.5 MB for a tgz file with the bin/ and lib/ subdirectories of the build directory (I haven't looked at the install yet, I need to make sure we're not installing e.g. libbinaryen.a in the SDK package). But I thought it wasn't outrageous on top of the full SDK package. On windows we are installing static version of both Binaryen and LLVM (because neither supports that mode on Windows) and nobody has complained about the size AFAIK. (LTO will actually be a size improvement on windows because it shrinks all of the binaries).
And yes I agree that 17% is likely an understatement for exactly the reason you said. But I figured it was enough to justify doing this, so I went ahead :)

sbc100:
Is the reason for the speedup because each individual tool then gets LTO'd with libbinaryen code?

We have customers that care a lot about speed of wasm-emscripten-finalize so anything that improves that is most likely a good idea. 

afterthefact lgtm.

PR #689, opened 2021-03-05
Include clang-resource-headers in the install targets for the LTO build

dschuff:
These are all the headers in lib/clang/<version> and are necessary.
Also remove unneeded symlinks/copies of lld in the install package.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -848,7 +848,8 @@ def CopyLLVMTools(build_dir, prefix=''):
                           'clang-func-mapping', 'clang-import-test',
                           'clang-offload-bundler', 'clang-refactor',
                           'clang-rename', 'clang-scan-deps', 'libclang.dll',
-                          'lld-link', 'ld.lld', 'lld64.lld', 'llvm-lib'):
+                          'lld-link', 'ld.lld', 'ld64.lld', 'llvm-lib',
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -917,7 +918,7 @@ def LLVM(build_dir):
         targets = ['clang', 'lld', 'llvm-ar', 'llvm-addr2line', 'llvm-cxxfilt',
                    'llvm-dwarfdump', 'llvm-dwp', 'llvm-nm', 'llvm-objcopy',
                    'llvm-objdump', 'llvm-ranlib', 'llvm-readobj', 'llvm-size',
-                   'llvm-strings', 'llvm-symbolizer']
+                   'llvm-strings', 'llvm-symbolizer', 'clang-resource-headers']
```
sbc100, line 921: Why do we have this list for lto but not for the normal build?

sbc100 (approved):

src/build.py
dschuff: Because the LTO build uses the `distribution` ninja build target instead of the default `all` target (see the next line, 922). We could (and maybe should) do this for the non-LTO build as well, but I figured I'd wait until the LTO build was stable before trying to merge them.

PR #690, opened 2021-03-05
Build binaryen with tools only

dschuff:
emsdk doesn't need to support linking programs against its libbinaryen.


sbc100 (approved):

sbc100:
What else might be installed?

dschuff:
This is basically just about libbinaryen.a (see https://github.com/WebAssembly/binaryen/pull/3654)

kripken (approved):

PR #691, opened 2021-03-10
Build with LTO automatically

dschuff:
Add a new mode to "automatically" build with LTO.
Specifically when the most recent commit in emscripten-releases modifies the
emscripten-version.txt file, build with LTO.
The use-lto flag can be set to "auto" for this behavior, or true or false.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -48,6 +48,7 @@ JSVU_OUT_DIR = os.path.expanduser(os.path.join('~', '.jsvu'))
 CMAKE_TOOLCHAIN_FILE = 'Wasi.cmake'
 
 EMSCRIPTEN_CONFIG_UPSTREAM = 'emscripten_config_upstream'
+EMSCRIPTEN_VERSION_FILE = 'emscripten-version.txt'
 
 # Avoid flakes: use cached repositories to avoid relying on external network.
 GIT_MIRROR_BASE = 'https://chromium.googlesource.com/'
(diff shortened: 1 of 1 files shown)


kripken:
> Specifically when the most recent commit in emscripten-releases modifies the emscripten-version.txt file, build with LTO.

This isn't the version that we release to users, however? We update that file in emscripten after we pick an emscripten-releases version whose existing CI results look good.

dschuff:
> > Specifically when the most recent commit in emscripten-releases modifies the emscripten-version.txt file, build with LTO.
> 
> This isn't the version that we release to users, however? We update that file in emscripten after we pick an emscripten-releases version whose existing CI results look good.

Today we do that. But the one in emscripten is not the one that propagates to emsdk users (I think? Because we don't update emscripten-version.txt in emscripten until after the tag, right?).

One idea that I proposed in the doc is to use emscripten-version.txt from the emscripten-releases package instead of from the git repo or created by emsdk itself. We can talk about that in the doc or separately though, because this PR doesn't actually do that; all I want from this PR is *some* way to trigger an LTO build that will be archived, so we can test it end-to-end.

kripken (approved):
Ok, I see, then this would be part of a change to `emscripten-releases.txt` then. sgtm to land for testing purposes as we figure that out (doesn't look like this could break anything meanwhile).

PR #692, opened 2021-03-20
Change emscripten and llvm-test-suite's remote branch

aheejin:
These two repos changed their default branch to `main` recently.


dschuff:
+cc @sbc100 
This change looks fine (since as you say upstream has switched to main, and the master branches are going away so there's no point in keeping this code as-is). 
But also it's only the wasm-stat.us waterfall that uses this (since emscripten-releases uses gclient's DEPS mechanism), and that waterfall has been red for a long time. I'm wondering if we should just shut it down if nobody is going to use it.

dschuff (approved):

src/build.py
```diff
@@ -690,9 +690,11 @@ def AllSources():
                LLVM_GIT_BASE + 'llvm-project.git',
                checkout=RemoteBranch('main')),
         Source('llvm-test-suite', GetSrcDir('llvm-test-suite'),
-               LLVM_GIT_BASE + 'llvm-test-suite.git'),
+               LLVM_GIT_BASE + 'llvm-test-suite.git',
+               checkout=RemoteBranch('main')),
```
dschuff, line 694: I guess we could even make 'main' the default now instead of 'master', and then I think it would only be waterfall and wabt that would need `RemoteBranch` (and those we should probably just fix anyway)

sbc100: Yeah I think its time to change the default.. although can can be a followup if you prefer.

sbc100 (approved):

src/build.py
sbc100: I just renamed wabt's branch to use `main` :)

dschuff:
followup is fine. this change is still correct and good on its own.

PR #694, opened 2021-03-23
Trigger LTO builds on locally-modified version file

dschuff:
Currently LTO builds are only triggered when the most recent commit modifies
the version file. But on try jobs, the files changed by the CL are not
committed; they are just modified in the index. So this change causes
LTO builds to also be triggered when the version file is modified
according to 'git status'.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -539,10 +539,22 @@ class Source(object):
 
 
 def RevisionModifiesFile(f):
-    # TODO: There's probably some nice single-command way to do this.
+    """Return True if the file f is modified in the index, working tree, or
+    HEAD commit."""
     if not os.path.isfile(f):
         return False
     cwd = os.path.dirname(f)
+    # If the file is modified in the index or working tree, then return true.
+    # This happens on trybots.
+    status = proc.check_output(['git', 'status', '--porcelain', f],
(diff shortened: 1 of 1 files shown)


src/build.py
```diff
@@ -539,10 +539,22 @@ def PrintGitStatus(self):
...
-    # TODO: There's probably some nice single-command way to do this.
+    """Return True if the file f is modified in the index, working tree, or
+    HEAD commit."""
     if not os.path.isfile(f):
         return False
     cwd = os.path.dirname(f)
+    # If the file is modified in the index or working tree, then return true.
+    # This happens on trybots.
+    status = proc.check_output(['git', 'status', '--porcelain', f],
+                               cwd=cwd).strip()
```
sbc100, line 550: Why `--porcelain`

sbc100 (approved):

src/build.py
dschuff: That flag makes 'git status' a stable parseable output format. in git terminology, 'plumbing' commands are meant to be used by other tools (and have stable outputs), whereas 'porcelain' commands are meant to be used by users, so they have pretty but unstable outputs. I have no idea why the flag to make `git status` work like a plumbing command is `--porcelain`.

PR #695, opened 2021-03-25
Suppress optional dependencies when installing native closure compiler

dschuff:
Without this, the 'npm install' step installs a *second* copy
of the native google-closure-compiler-{os} package (as a dependency inside
the google-closure-compiler directory) in addition to the one installed
at the top level by the primary command.

### DIFF
--- a/src/build.py
+++ b/src/build.py
@@ -1157,7 +1157,8 @@ def InstallEmscripten():
         native = 'google-closure-compiler-windows'
     elif IsLinux():
         native = 'google-closure-compiler-linux'
-    proc.check_call(['npm', 'install', '--no-optional', native], cwd=em_install_dir)
+    proc.check_call(['npm', 'install', '--no-optional', native],
+                    cwd=em_install_dir)
 
     version_file = GetSrcDir(EMSCRIPTEN_VERSION_FILE)
     if os.path.isfile(version_file):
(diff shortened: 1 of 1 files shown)


sbc100:
And does the install version get found an used now?

sbc100 (approved):
Weird...

kripken:
The second copy was of a native version?

[The docs](https://www.npmjs.com/package/google-closure-compiler) suggest that the native version should arrive as an optional dep, and in no other way ("Native Binary Version" section on that page). Maybe that's out of date, but otherwise it sounds like this change would prevent us from getting a native version?

dschuff:
the second copy was a native version.  `npm install google-closure-compiler-linux` was installing 2 different copies with different versions. With this change, it only installs one.

kripken:
So there is one java version, and two native versions installed? Or do you mean two native versions and nothing else?

Maybe we should file a bug on closure compiler? I'm a little uncomfortable working around it, as it might be fixed tomorrow and the docs could become accurate again, and we'd silently lose all the speed of the native version.

sbc100:
> So there is one java version, and two native versions installed? Or do you mean two native versions and nothing else?
> 
> Maybe we should file a bug on closure compiler? I'm a little uncomfortable working around it, as it might be fixed tomorrow and the docs could become accurate again, and we'd silently lose all the speed of the native version.

We are already working around two different upstream closure compiler bugs here... see the command above that strange codepath we are taking here.   This change doesn't make it any worse or better IMHO.

kripken (approved):
Ok, I see, sgtm...

